PR #16, opened 2020-05-25
adding RPC method GetChainID to the access protobuf to retreive chain ID

vishalchangraniaxiom:
Closes: 

## Description

This change adds a new RPC method GetChainID to the Access protocol. The chain id is used to distinguish between flow mainnet and testnet.
______

For contributor use:


docs/access-api-spec.md
```diff
@@ -626,6 +628,38 @@ The event results are grouped by block, with each group specifying a block ID an
   ```
 </details>
 
+---
+
+### Chain ID
+
+ChainID helps identify the Flow network (mainnet versus testnet). The following method can be used to query for the Chain ID
```
vishalchangraniaxiom, line 635: do you think it add values if we also report `uptime`? I could extend the GetChainID method to report more details about the network.

psiemens: Yeah, I was thinking it might be worthwhile to generalize this a bit. 

Maybe `GetNetworkParameters`?

`chain_id` is just one of the parameters IMO. We might want to return information about the epoch, gas prices, etc in the future.

```
message GetNetworkParametersResponse {
    string chain_id = 1;
}
```

vishalchangraniaxiom: how about `GetChainDetails`?

docs/access-api-spec.md
```diff
@@ -855,3 +888,17 @@ message Event {
 | transaction_index | Zero-based index of the transaction within the block |
 | event_index       | Zero-based index of the event within the transaction |
 | payload           | Event fields encoded as [JSON-Cadence values](/docs/json-cadence-spec.md)|
+
+### Network Parameters
+
+Network parameters provide information about the flow network. Currently, it only includes chain-id
```
psiemens, line 894:
```suggestion
Network parameters provide information about the Flow network. Currently, it only includes chain ID.
```

docs/access-api-spec.md
```diff
@@ -855,3 +888,17 @@ message Event {
...
+
+```
+message NetworkParameters {
+	string chain_id = 1;
+}
+```
+
+| Field             | Description    |
+| ------------------|----------------|
+| chain_id          | ChainID helps identify the Flow network. It can be one of `flow-mainnet`, `flow-testnet` or `flow-emulator` |
```
psiemens, line 904:
```suggestion
| chain_id          | Chain ID identifies the Flow network and can be one of `flow-mainnet`, `flow-testnet` or `flow-emulator` |
```

protobuf/flow/entities/network_parameters.proto
```diff
@@ -0,0 +1,7 @@
+syntax = "proto3";
+
+package entities;
+
+message NetworkParameters {
```
psiemens, line 5: Sorry to be difficult, but I don't think `NetworkParameters` needs to be its own entity. Up until now we've been reserving entities for "items that we hash" and they are directly correlated with the models that we store.

I'd prefer to just have each network param be a field in `GetNetworkParametersResponse`.

The table you added above is great, though! I would just move it up to the documentation for `GetNetworkParameters`

vishalchangraniaxiom: ok will change it

vishalchangraniaxiom: > Sorry to be difficult, but I don't think `NetworkParameters` needs to be its own entity. Up until now we've been reserving entities for "items that we hash" and they are directly correlated with the models that we store.
> 
> I'd prefer to just have each network param be a field in `GetNetworkParametersResponse`.
> 
> The table you added above is great, though! I would just move it up to the documentation for `GetNetworkParameters`

and you are not being difficult :)

PR #18, opened 2020-05-30
Add timestamp to block header

philipstanislaus:
Required by https://github.com/dapperlabs/flow-go/issues/3271 and https://github.com/dapperlabs/flow-go/issues/2255


philipstanislaus:
Thanks! We are using `google.protobuf.Timestamp` already in the block protobuf, so that should be no issue @turbolent . Merging now.

PR #23, opened 2020-06-09
Add expired status

Kay-Zee:
## Description

Add an `Expired` status for transactions


protobuf/flow/entities/transaction.proto
```diff
@@ -8,6 +8,7 @@ enum TransactionStatus {
   FINALIZED = 2;
   EXECUTED = 3;
   SEALED = 4;
+  EXPIRED = 5;
```
vishalchangrani, line 11: are we also making equivalent change in the access node code in flow-go to support this status?

Kay-Zee: https://github.com/dapperlabs/flow-go/issues/4023

Kay-Zee: Having the status here doesn't negatively affect the node code in flow-go i think. e.g. EXECUTED status isn't yet being returned either, which is part of the above linked issue

PR #26, opened 2020-06-16
adding GetAccountAtBlockHeight API definition

vishalchangrani:
Closes: #4110

## Description

This PR adds a new API call - `GetAccountAtBlockHeight`

______

For contributor use:


docs/access-api-spec.md
```diff
@@ -439,13 +440,42 @@ rpc GetAccount(GetAccountRequest) returns (GetAccountResponse)
   ```
 </details>
 
+#### GetAccountAtBlockHeight
+
+`GetAccountAtBlockHeight` gets an [account](#account) by address at the given block height
```
psiemens, line 445:
```suggestion
`GetAccountAtBlockHeight` gets an [account](#accounts) by address at the given block height.
```

protobuf/flow/access/access.proto
```diff
@@ -49,8 +49,10 @@ service AccessAPI {
 
   // Accounts
 
-  // GetAccount gets an account by address.
+  // GetAccount gets an account by address from the latest sealed execution state
```
psiemens, line 52:
```suggestion
  // GetAccount gets an account by address from the latest sealed execution state.
```

protobuf/flow/access/access.proto
```diff
@@ -49,8 +49,10 @@ service AccessAPI {
...
   rpc GetAccount (GetAccountRequest) returns (GetAccountResponse);
```
psiemens, line 53:
```suggestion
  rpc GetAccountAtLatestBlock (GetAccountRequest) returns (GetAccountResponse);
```

How do you feel about updating this to be consistent with the script methods below?

protobuf/flow/access/access.proto
```diff
@@ -160,6 +162,15 @@ message GetAccountResponse {
...
 }
 
+message GetAccountAtBlockHeightRequest {
+  bytes address = 1;
+  uint64 block_height = 2;
+}
+
+message GetAccountAtBlockHeightResponse {
+  entities.Account account = 1;
+}
```
psiemens, line 172: Just like with the `GetBlock` methods, all of the `GetAccount` methods can share a common `AccountResponse`

docs/access-api-spec.md
```diff
@@ -439,13 +440,42 @@ rpc GetAccount(GetAccountRequest) returns (GetAccountResponse)
...
+  ```
+</details>
 ---
 
 ### Scripts
 
 #### ExecuteScriptAtLatestBlock
 
-`ExecuteScriptAtLatestBlock` executes a read-only Cadance script against the latest sealed execution state.
+`ExecuteScriptAtLatestBlock` executes a read-only Cadence script against the latest sealed execution state.
```
psiemens, line 478: 👍

protobuf/flow/access/access.proto
vishalchangrani: I thought about that, then I felt that it would be a breaking change for the NBA folks.

vishalchangrani:
> At latest always means sealed in our case, i assume? Do we want to add the option to do any state query at finalized? I'm guessing that's usually not desire, but for some applications that are trying to be optimistic, it's possible,

`GetAccount` could have an `is_sealed` option as part of the request and in that case the access node gets the block id of the sealed or the final block. But what should be done for `GetAccountAtBlockHeight`, how to determine that the block at height is sealed or finalized? Should I check the height against the height of the latest sealed block?

vishalchangrani:
> At latest always means sealed in our case, i assume? Do we want to add the option to do any state query at finalized? I'm guessing that's usually not desire, but for some applications that are trying to be optimistic, it's possible,

I have created an issue for this - https://github.com/dapperlabs/flow-go/issues/4150. I might have to think this through a little.

PR #30, opened 2020-07-14
Access API: Add revoked flag to AccountKey

psiemens:
## Description

This PR updates the Access API to specify whether or not an account key has been revoked from an account.

This change is necessary in order to support an update to the way we are handling key deletion. Rather than deleting a key from an account entirely, we will now be marking it as revoked.


docs/access-api-spec.md
```diff
@@ -900,6 +901,7 @@ message AccountKey {
 | hash_algo       | [Hash algorithm](/docs/accounts-and-keys.md#supported-signature--hash-algorithms) |
 | weight          | [Weight assigned to the key](/docs/accounts-and-keys.md#weighted-keys) |
 | sequence_number | [Sequence number for the key](/docs/accounts-and-keys.md#sequence-numbers) |
+| revoked         | Flag showing whether or not the key has been revoked |
```
vishalchangrani, line 904:
```suggestion
| revoked         | Flag indicating whether or not the key has been revoked |
```

PR #33, opened 2020-07-22
Add back GetAccount method for backwards compatibility

psiemens:
This PR reintroduces `GetAccount` as a temporary measure to maintain backwards compatibility with clients.


protobuf/flow/access/access.proto
```diff
@@ -49,6 +49,8 @@ service AccessAPI {
 
   // Accounts
 
+  // GetAccount is an alias for GetAccountAtLatestBlock.
```
vishalchangrani, line 52: shouldn't we also mention here that `GetAccount` is now been deprecated?

docs/access-api-spec.md
```diff
@@ -410,6 +411,36 @@ rpc GetTransactionResult (GetTransactionRequest) returns (TransactionResultRespo
 
 ### Accounts
 
+#### GetAccount
+
+`GetAccount` gets an [account](#account) by address at the latest sealed block.
+
+Note: this function behaves identically to `GetAccountAtLatestBlock` and will be deprecated in a future version.
```
vishalchangrani, line 418: isn't it already deprecated?

psiemens: Ah true, I'll update both comments to reflect this. Thanks!

PR #39, opened 2020-08-26
Document the transaction status

turbolent:
The transaction status is AFAICS currently not documented anywhere except for the whitepapers.

We should at least find concise descriptions for all statuses. I'm not really sure what unknown, pending, and expired mean.


psiemens (approved):
Thanks for adding this! I filled in the remaining statuses

Issue #40, opened 2020-09-10
BlockEvents to return timestamp


bjartek:
### Issue To Be Solved

When fetching events you might want to show when they happend. Right now you get the blockId back and you can do a seperate request to fetch the acutal timestamp for that. However if fetching events for a large range that will generate a lot of uneccesary requests. 

### (Optional): Suggest A Solution

BlockEvents returns a Timestamp field

### (Optional): Context

Working on a flow.Event -> Discord webhook script

bjartek:
Thanks!

PR #41, opened 2020-09-21
Add BlockTimestamp to EventsResponse

janezpodhostnik:
ref: #40

## Description

Added BlockTimestamp to the access node  EventsResponse. This allows you to know when the event was published without re-requesting the block by block id.


psiemens (commented):
Looks good! Just needs the small field ordering change we discussed

psiemens (approved):
Looks good to me! 🚢

PR #48, opened 2020-09-25
Add support for highlighting inline code

turbolent:
This PR enables inline code which is prefixed with a language tag (identifier + `•`) to be highlighted:

```
`cadence•fun test() {}`
````

This is not limited to Cadence code, but very useful in the Cadence documentation.

### DIFF
--- a/docs/plugins/gatsby-theme-flow/gatsby-config.js
+++ b/docs/plugins/gatsby-theme-flow/gatsby-config.js
@@ -64,6 +64,7 @@ module.exports = ({
           'solidity': 'source.solidity.security',
           'javascript': 'source.js',
           'shell': 'source.shell',
+          'sh': 'source.shell',
           'go': 'source.go',
           'js': 'source.js',
           'json': 'source.json',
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: docs/plugins/gatsby-theme-flow/gatsby-config.js [hunk 1]
```diff
@@ -64,6 +64,7 @@ module.exports = ({
           'solidity': 'source.solidity.security',
           'javascript': 'source.js',
           'shell': 'source.shell',
+          'sh': 'source.shell',
```
psiemens, line 67: 👍

psiemens: Can we add `protobuf` too? Not sure if `source.protobuf` will just work?

psiemens (approved):

docs/plugins/gatsby-theme-flow/gatsby-config.js
turbolent: Good point! I saw it's missing. I'll add in a separate PR

PR #53, opened 2020-09-26
Update grammars for syntax highlighting, list third-party licenses

turbolent:
We've copied grammars from other repositories – update them and properly give credits by listing third-party licenses.

I've removed the grammar for Solidity, because it unfortunately does not have any license.


psiemens (approved):
👌

PR #67, opened 2020-10-02
Added contracts on the access API Account entity to support multiple contracts per account

janezpodhostnik:
References: https://github.com/dapperlabs/flow-go/issues/4806

## Description

After the switch to multiple contracts per account, the "code" is actually a map of **contract name**: string -> **contract code**: byte array.

In order to keep backwards compatibility account has been changed to:

```proto
message Account {
  bytes address = 1;
  uint64 balance = 2;
  bytes code = 3;
  repeated AccountKey keys = 4;
  map<string, bytes> contracts = 5;
}
```

where the code element will always be empty.


docs/content/protocol/access-api.md
```diff
@@ -850,15 +850,17 @@ message Account {
...
+  map<string, bytes> contracts
 }
 ```
 
 | Field        | Description |
 |--------------|-------------|
 | address      | A unique account identifier |
 | balance      | The account balance |
-| code         | The code deployed to this account |
+| code         | The code deployed to this account (deprecated, use contracts instead) |
```
turbolent, line 861: 👌

turbolent (approved):
👍

Issue #69, opened 2020-10-05
Search should should not point you to docs.onflow.org on vercel (and local) deployments


janezpodhostnik:
### Problem

Searching the docs during local development (or on vercel deployment) incorrectly points you to the actual docs page.

Even though it will be incorrectly indexed for new or edited files (because the indexing is from https://docs.onflow.org/). I still thing you should be able to navigate using the search, to make things easier.

### Steps to Reproduce 

1. go to any vercel deployment. 
2. search for anything 
3. you are redirected to https://docs.onflow.org/

### Acceptance Criteria

Don't get redirected to https://docs.onflow.org/ when searching. Instead go to the current deployments pages.

PR #70, opened 2020-10-05
Replace origin on search results

janezpodhostnik:
Closes: #69 

## Description

Replace search result origin so search results point to local (or vercel) deployment


psiemens (approved):
Nice catch! Thanks for fixing this

PR #71, opened 2020-10-05
Add staking transaction doc. Fix typo in transactions doc.

robmyers:
Transient features branch for existing branch.

### DIFF
--- a/docs/content/token/staking2/stakers.mdx
+++ b/docs/content/token/staking2/stakers.mdx
@@ -32,45 +32,98 @@ _Note: The corresponding **node operator** steps are outlined in a
 
 ## 1. Fetch node operator information
 
-TODO
-- Provide `operatorAddress, nodeID`
-- Uses `TH.04` script
+To fetch the node operator information for an address's staking proxy, the token holder must use use script TH.04:
+
+`stakingProxy/sh_get_node_info.cdc`
+
+Script arguments:
+
+| Argument | Type | Purpose |
+| account  | Address | Specifies the address to get the staking proxy's node info node info for. |
+
+Script returns:
+
+The `StakingProxy.NodeInfo` struct containing the details for the node. If no NodeInfo is available, the script will fail.
 
 ## 2. Register node & grant staking access to the node operator
 
-TODO
-- Provide `amount` to be staked
-- Provide `operatorAddress, nodeID` to stake to
-- Uses `TH.05` transaction
+To register the node as the token holder account's staker, the token holder must use  use transaction TH.05:
+
+`stakingProxy/sh_register_node.cdc`
+
+Transaction arguments:
+
+| Argument | Type | Purpose |
+| address  | Address | The node operator address. |
+| nodeID   | String  | The `id` from the NodeInfo. |
+| amount   | UFix64  | The number of FLOW tokens to stake. |
 
 # Staking Actions
 
 ## Stake New Tokens
 
-TODO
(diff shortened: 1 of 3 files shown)


docs/content/token/staking2/stakers.mdx
```diff
@@ -32,45 +32,105 @@ _Note: The corresponding **node operator** steps are outlined in a
 
 ## 1. Fetch node operator information
 
-TODO
-- Provide `operatorAddress, nodeID`
-- Uses `TH.04` script
+To fetch the node operator information for an address's staking proxy, the token holder must use use script TH.04:
```
psiemens, line 35: It'd be good to add a bit of context as to why this step is required. Something like this:

> After a node operator has agreed to run a node for you, they will create a `NodeInfo` object holding the IP address and public keys for your new node. This `NodeInfo` object will be stored in an account owned by the node operator.
> 
> The node operator should provide you with their account `address` and the `nodeID` for your new node. These parameters can be passed into the `get_node_info` script (`TH.04`) to confirm the node information.
>
> If the node information is correct, continue to step 2.

docs/content/token/staking2/stakers.mdx
```diff
@@ -32,45 +32,105 @@ _Note: The corresponding **node operator** steps are outlined in a
...
+
+`stakingProxy/sh_get_node_info.cdc`
+
+Script arguments:
+
+| Argument     | Type    | Purpose |
+|--------------|---------|---------|
+| **account**  | Address | Specifies the address to get the staking proxy's node info node info for. |
```
psiemens, line 43:
```suggestion
| **account**  | Address | The address of the node operator's account. |
```

IMO it's more clear to just say this is the node operator's account.

docs/content/token/staking2/stakers.mdx
```diff
@@ -32,45 +32,105 @@ _Note: The corresponding **node operator** steps are outlined in a
...
+
+The `StakingProxy.NodeInfo` struct containing the details for the node. If no NodeInfo is available, the script will fail.
 
 ## 2. Register node & grant staking access to the node operator
 
-TODO
-- Provide `amount` to be staked
-- Provide `operatorAddress, nodeID` to stake to
-- Uses `TH.05` transaction
+To register the node as the token holder account's staker, the token holder must use  use transaction TH.05:
```
psiemens, line 51: Can you make it so all of the transaction IDs (e.g. `TH.05`) link to the relevant section of `transactions.mdx`? I know you can't link directly to the line in the table, but linking to the whole table is enough I think

psiemens (commented):
Looking good!

docs/content/token/staking2/stakers.mdx
robmyers: Added. Thank you!

docs/content/token/staking2/stakers.mdx
robmyers: I've done this in rob/docs/staking-as-token-holder .

docs/content/token/staking2/stakers.mdx
robmyers: Wow yes. I've changed this in rob/docs/staking-as-token-holder .

docs/content/token/staking2/stakers.mdx
robmyers: Done.

psiemens (approved):

PR #74, opened 2020-10-05
Add table formatting

MaxStalker:
Fix table formatting in mdx


robmyers (approved):
Thank you for catching and fixing this!

PR #77, opened 2020-10-06
editing the staking doc

joshuahannan:
## Description

add more detail to the staking doc and change file names


docs/content/token/staking2/stakers.mdx
```diff
@@ -50,7 +53,7 @@ Script arguments:
 
 Script returns:
 
-The `StakingProxy.NodeInfo` struct containing the details for the node. If no NodeInfo is available, the script will fail.
```
robmyers, line 53: This is "the struct in particular that contains the details for this specific node" rather than "the struct of this type with these properties" but I prefer the revision to the latter. :-)

robmyers (approved):
Thank you for catching these!

PR #80, opened 2020-10-06
Adds account creation doc

joshuahannan:
## Description

Adds instructions for custody providers to create locked token accounts for their users.
Probably needs more details and transaction labels like we have for the other transactions.e Github PR explorer


robmyers (approved):
Looks good!

psiemens (approved):

PR #135, opened 2020-10-23
Update delivery.mdx

laynelafrance:


caseyturnerarrington (approved):
@laynelafrance looks good. I just think we should remove 'located [here]', as there is no link there.

PR #136, opened 2020-10-23
Add addresses for other contracts to the docs

robmyers:
But just those used in the core contracts transactions (so no NFT contract for example).

I've linked this page from the upcoming FAQ.


psiemens (approved):
Great idea!

PR #145, opened 2020-11-03
Clarify account code fields

turbolent:
Just looking at the Protobuf file, the `bytes` type could be assumed to be byte code. Clarify and prevent the misconception.


psiemens (approved):
👌

PR #151, opened 2020-11-05
Remove "docset" dropdown menu from all pages

psiemens:
## Description

The dropdown menu has been bothersome for a while now -- it's a bit glitchy and doesn't do a great job as a navigation tool. I'd like to remove it for now and start to work on a better solution.

The site is just as usable IMO -- readers will need to navigate back to the main page to browse other sections, but I think that's OK.


joshuahannan (approved):
Looks good to me. It isn't totally clear to me where I can navigate if I've went into one of the sections and want to get out without pressing the back button. I see that clicking on the Flow docs logo takes you back to the homepage but it might be good if that was clearer

PR #153, opened 2020-11-05
Revamp staking docs for new content structure

psiemens:
The old staking documentation assumed that all stakers would be using a third-party operator via the `StakingProxy` contract. This PR attempts to adjust the documentation to better reflect the general case in which the token holder does not relinquish control of their tokens.

The third-party operator case has been moved to a separate page: [Stake with a Third-party Operator](https://flow-docs-git-staking-docs-update.onflow.vercel.app/staking/third-party-operator/)

### DIFF
--- a/docs/content/staking/delegation-transactions.mdx
+++ b/docs/content/staking/delegation-transactions.mdx
@@ -1,22 +1,28 @@
 ---
-title: How to Delegate Locked FLOW
-sidebar_title: Delegation Transactions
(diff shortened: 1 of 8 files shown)


joshuahannan (approved):
Looks good, but I still think that we should still include "Locked FLOW" in the title of these so it is clear

psiemens:
> I still think that we should still include "Locked FLOW" in the title of these so it is clear

Good call, I'll update

PR #155, opened 2020-11-06
Changes request to register

joshuahannan:
## Description

* Changed description to node registration instead of staking request
* Documented that unstake requests will uncommit tokens if there are any.


psiemens (approved):
Nice, this is much clearer 👍

PR #200, opened 2020-11-26
FLIP-0200: Application of BIP 44 in Flow Wallets

psiemens:
This PR proposes FLIP 200: **Application of BIP 44 in Flow Wallets**.


flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Coin Type
+
+Coin type is always the constant 539' (0x8000021b), 
+the unique identifier for the Flow protocol, as defined in 
+[SatoshiLabs Improvement Proposal 44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) (SLIP 44).
+
+#### Account
+
+Account is the index of a distinct account address owned by the wallet user.
```
Kay-Zee, line 60: This index also has no relation to the `Address state index` used to generate the Flow Addresses, correct? Is that worth noting? It's something i keep thinking about, but am likely an outlier.

flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,196 @@
...
+by 1.
+
+This index should not be confused with the on-chain account key index.
+
+### Account Discovery
+
+Wallet developers should use a modified version of the original account discovery
+procedure [described in BIP 44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery).
+
+1. Derive the first account's node (index = 0).
```
Kay-Zee, line 95: I feel like the `node` terminology is first introduced here. May be worth more definition before use? Or is it self explanatory?

flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,196 @@
...
+2. Scan for addresses by checking each public key index against the public key registry,
+respecting the [gap limit described in BIP 44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit).
+3. If no address is found in the registry, stop discovery.
+4. If an address is found, query the chain to fetch the account details.
+5. If no account is found<sup>1</sup>, skip this index and go to step 1.
+6. If an account is found, add it to the user's wallet and go to step 1.
```
Kay-Zee, line 101: Is there an increment step missing? Isn't step 1 going to use index=0 again?

flips/20201125-bip-44-multi-account.md
psiemens: Great point. There's no relation, I'll make that more clear

flips/20201125-bip-44-multi-account.md
psiemens: "Node" is a term carried over from BIP 44, but good point -- it isn't explained here. I'll add more clarity to this section

flips/20201125-bip-44-multi-account.md
psiemens: Yep good catch, I'll add!

flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,196 @@
...
+also be 0, 1, 2, resulting in a clear mapping from each on-chain key 
+to its corresponding derivation path.
+
+The authors of this proposal chose to omit this restriction after considering that,
+in many cases, a user may add additional public keys from other sources, thus breaking
+the perfect alignment between on-chain and key path indices. Even without outside keys,
+it is still possible for keys to be added out of order.
+
+However, wallet developers or users may still incorrectly assume that the path 
+index of a public key always matches its on-chain account key index.
```
robmyers, line 141:
```suggestion
Wallet developers or users should therefore not rely on the false assumption that the path 
index of a public key always matches its on-chain account key index, as doing so
will cause confusion and errors.
```

Just to spell out the consequences.

psiemens: This is great, thank you!

flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,196 @@
...
+
+In particular, this proposal impacts wallet software that is compatible
+with the Flow Ledger application, such as Flow Port.
+
+## Related Issues
+
+### Mapping Public Keys to Addresses
+
+As mentioned in the [drawbacks](#drawbacks) section, this proposal necessitates 
+a public key registry that maps public keys to account addresses.
```
robmyers, line 182:
```suggestion
a public key registry that maps public keys to account addresses. The information required
to associate Flow addresses with their keys is broadcast when keys are added to Flow
accounts both during and after account creation. Making this information searchable
requires indexing this information rather than requiring any additional disclosure on the
part of Flow account owners.
```

flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,197 @@
...
+#### Coin Type
+
+Coin type is always the constant 539' (0x8000021b), 
+the unique identifier for the Flow protocol, as defined in 
+[SatoshiLabs Improvement Proposal 44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md) (SLIP 44).
+
+#### Account
+
+Account is the index of a distinct account address owned by the wallet user. 
+No two account indices should be used to generate child public keys that are assigned
```
tarakby, line 61:
```suggestion
No two distinct account indices 
```
?

flips/20201125-bip-44-multi-account.md
```diff
@@ -0,0 +1,323 @@
...
+### Account Discovery
+
+Wallet developers should use a modified version of the original account
+discovery procedure [described in BIP
+44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery).
+
+Modified constants from BIP 44 include:
+
+The key index gap limit should be `5`.
+The account index gap limit should be `5`.
```
PeterBenc, line 143: Why `5` ? Seems like a lot. Does this mean a wallet should derive 2(two curves) * 5(key_gap) * 5(account_gap) keys in order to say that a wallet is not used? In other words, should the wallet derive 10 keys, just to say the account is not used? Am I getting this wrong? What is the use case for such a big gap limit?

Since keys and accounts should be used in order (meaning the first unused should be picked when creating new account/key) why not gap limit `1`?.

JeffreyDoyle: Hey @PeterBenc , after consulting with the team, we've concluded that the key gap limit and account gap limit parameters should be rather flexible. Developers who wish to use a smaller key or address gap limit should be able to do so if they choose - developers should use their best judgement when choosing a key gap limit and account gap limit.

PeterBenc: Thanks for considering this, makes sense, at least for now we are gonna use account_gap_limit 1, and check only the first key (so I guess key_gap_limit 0) since right now, basically nobody is deriving these keys anyway and we can increase the gap limit later if needed. 

Just genuinely interested, I suppose the 5 and 2 (for key account gap respectively) should have some recommending purpose? Since it's supposed to be flexible, what is the reason for defining these?

bluesign: @PeterBenc
In BTC for example addresses are bound to keys, so if you have HD wallet you can have unlimited addresses.

so lets say my wallet pre generates N addresses for me, and I use the Nth to receive some coins. 

After I restore my wallet to another computer, you need to search N addresses (gap = N ) otherwise I see my funds are missing and panic 

i dont know why we need in Flow , I need to reread to FLIP

tarakby:
> Just genuinely interested, I suppose the 5 and 2 (for key account gap respectively) should have some recommending purpose? Since it's supposed to be flexible, what is the reason for defining these?

@PeterBenc Considering 5 and 2, the number of key queries required to reach a gap is 2*2*5 = 20 (curves * key gap * account gap). This matches the number of queries recommended by BIP 44 [https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit]. 
This is only an estimation as it is not known if a query to the Flow key registry is slower or faster than an on-chain query. The complexity (20) is equivalent, but numbers can be further adjusted by the wallet implementation.

pgebheim (approved):
Proposal has been well-argued and has full implementations.

Approving for merge.

PR #208, opened 2020-11-27
adds lease section

joshuahannan:


caseyturnerarrington:
@joshuahannan for this section, do we want to say something around the fact that a user will potentially need to create 2 separate accounts (to cover the case where a person is trying to stake with their leased flow amount, but that amount isn't enough to cover the min, so they need to delegate to it via another account)?

joshuahannan:
@caseyturnerarrington That currently isn't possible. They need to have all the tokens in one account in order to meet the minimum. Any of our node operators who don't have enough of their own tokens are being leased the minimum for their node and they just delegate their tokens normally

caseyturnerarrington:
@joshuahannan This is the reason why we reduced the Consensus minimum to being half the min it previously was :) (to handle this exact case). Ex:
- Person gets 250,000 lease and 250,000 token grant.
- They would: Stake with the 250k lease, it would be processed, then they would delegate the remaining 250 grant via a separate account. 

I think the way I said it above is confusing tho. It should be more like: 'If you are receiving a token lease and a token grant (or any other type of token allocation), you will need to create multiple accounts. The token lease needs to be in its own account.' (or something like that)

joshuahannan:
that makes sense. I was speaking about the minimum that is defined in the smart contract (250k), not the future minimum. I'll add your clarification

caseyturnerarrington:
@jordanschalm made some updates. want to take a peek?

jordanschalm:
@caseyturnerarrington I'm guessing that was meant to be tagging @joshuahannan 😅

caseyturnerarrington:
Gosh dangit @jordanschalm @joshuahannan moving too fast over here!!

joshuahannan:
Looks good to me! Want to give an approval @caseyturnerarrington ?

caseyturnerarrington (approved):

10thfloor (approved):

PR #226, opened 2020-12-01
Migrate all repo landing pages from github.

10thfloor:
Migrate the repo README content into the docs site for searchability.

New pages: 
- VSCode Introduction
- JS SDK Introduction
- Flow Emulator Introduction


psiemens (commented):
What if we just combined `/devtools` and `/sdks` into a single section called `/tools`?

Issue #214, opened 2020-11-30
Update Tutorials to work with latest version of Cadence


joshuahannan:
### Issue To Be Solved
Cadence has introduced some breaking changes and the tutorials need to be updated to work with those.

### (Optional): Suggest A Solution
* Go through each tutorial and test with the latest version of Cadence
* Update any out of date patterns or syntax
* For example, the marketplace example stores a reference, which needs to actually be a capability.

Tutorials are here: https://github.com/onflow/flow/tree/master/docs/content/tutorial/cadence

PR #247, opened 2020-12-09
updates tutorials with best practices and fixes

joshuahannan:
Closes: #214 

## Description

* Changed parts of the hello world tutorial to use optional chaining
* Updated marketplace tutorial to use a capability instead of a reference
______

For contributor use:

### DIFF
--- a/docs/content/tutorial/cadence/02-hello-world.mdx
+++ b/docs/content/tutorial/cadence/02-hello-world.mdx
@@ -499,7 +499,7 @@ let helloReference = capability!.borrow()
 This method creates the reference as the type we specified in `<>` in the `link` function. 
 Here we use the force-unwrap operator (`!`) because the capability is an optional. If the capability is `nil` the transaction will abort.
 While borrowing the reference, we use optional chaining because the borrowing of the reference could fail. 
-It the reference could be `nil` if the targeted storage slot is empty, is already borrowed, or if the requested type exceeds what is allowed by the capability.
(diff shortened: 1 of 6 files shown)


docs/content/tutorial/cadence/02-hello-world.mdx
```diff
@@ -471,22 +475,37 @@ First, we create a capability that is linked to the private `HelloAsset` object
...
 let helloReference = capability!.borrow()
+    ?? panic("Could not borrow a reference to the hello capability")
 ```
 
-This method creates the reference as the type we specified in `<>` in the `link` function. Here we use the force-unwrap operator (`!`) because the capability is an optional. If the capability is `nil` the transaction will abort.
-It also returns `nil` if the targeted storage slot is empty, is already borrowed, or if the requested type exceeds what is allowed by the capability.
+This method creates the reference as the type we specified in `<>` in the `link` function. 
+Here we use the force-unwrap operator (`!`) because the capability is an optional. If the capability is `nil` the transaction will abort.
+While borrowing the reference, we use optional chaining because the borrowing of the reference could fail. 
+It the reference could be `nil` if the targeted storage slot is empty, is already borrowed, or if the requested type exceeds what is allowed by the capability.
```
10thfloor, line 502:
```suggestion
The reference could be `nil` if the targeted storage slot is empty, is already borrowed, or if the requested type exceeds what is allowed by the capability.
```

10thfloor (changes requested):
Looks good. One small change.

10thfloor (approved):

PR #266, opened 2020-12-11
remove account setup from staking and add NFT page

joshuahannan:
* The locked/lease account setup was in both the staking section and token section. It should only be in the token section so I removed it from staking
* Added the testnet and mainnet addresses for the NFT standard interface
* Moved `LockedTokens` to Other Important Contracts because it is not a "core contract"


docs/content/protocol/core-contracts/non-fungible-token.mdx
```diff
@@ -0,0 +1,14 @@
+---
+title: Non-Fungible Token Contract Reference
+sidebar_title: Non-Fungible Token
+---
+
+The `NonFungibleToken` contract interface implements the Fungible Token Standard.
```
turbolent, line 6: Should this be?

```suggestion
The `NonFungibleToken` contract interface implements the Non-Fungible Token Standard. 
```

turbolent (approved):

PR #280, opened 2020-12-15
Fix contract deployment command

turbolent:
Use the new CLI flag `--contract` instead of the outdated `--code` flag


10thfloor:
Closed by #285

turbolent:
@10thfloor you missed the contract name in the contract flag

10thfloor:
@turbolent Added here: 8059bdd900

PR #292, opened 2020-12-16
Add missing redirects

10thfloor:
Adds `308` redirects for old links after updating page hierarchies.

Note: Redirects are 307 while testing, and will change to permanent  (308) once everything is tested.


psiemens (approved):
Looks good, thank you!

PR #324, opened 2021-01-19
FLIP for self-ejection (feature for core protocol)

AlexHentschel:
This PR proposes FLIP xyz: `Self-Ejection (feature for the core protocol)`


jordanschalm (commented):
I think this would be better implemented within the service account smart contracts. The node wishing to self-eject can submit their rejection request via a transaction to the appropriate service account contract. That contract can then validate the request and propagate the resulting identity table change to the consensus committee via a service event. 

This matches the flow of information we have for identity table changes during epoch transitions: nodes submit staking requests to a smart contract, the requests are validated and aggregated into an identity table change, then the identity table change is propagated to the consensus committee (and ultimately to the protocol state) via a service event. 

It is also simpler for a node operator wishing to submit an ejection request to submit a regular Flow transaction than to arrange to pass a message to consensus nodes directly (through the privileged protocol networking layer), since this would be a manual operation. 

For other identity table changes, for example as a result of slashing challenges, my mental model has been:
* consensus committee adjudicates a slashing challenge and includes it in a block
* the service account reads new slashing challenges included in each block and modifies stake/weight/ejected fields of the node accordingly and emits a service event containing these identity table changes
* the protocol state is updated when the service event is processed/sealed by the consensus committee

In other words, the service account decides when the identity table changes. The consensus committee can updates its version of the identity table upon receipt of messages from the service account. And the consensus committee can provide "hints" to the service account which it may use to decide upon identity table changes.

AlexHentschel:
Thanks for the feedback @jordanschalm. Your points make sense to me. 

It seems we have two different proposals for handling slashing challenges and node ejections. 
1. the one you describe: 
     - the core protocol only adjudicates slashing challenges
     - the punishment for protocol violations is handled by the smart contract
2. The proposal I described [here (notion)](https://www.notion.so/dapperlabs/Protocol-Violations-and-Slashing-fc1b37ab41274e91abfb9ce1cb3f8d54)
     > Generally, we distinguish between two types of protocol violations: 
     > - **felony:**  a *severe* protocol violation that is immediately punished by *ejecting* the respective node from the network and subsequent financial penalty.
     > Most protocol violations are felonies, including
     > • equivocation by Consensus nodes or Collector nodes
     > • publishing (EN) or approving (VN) a faulty Execution Result 
     > • sending invalid messages (such as blocks)
     > - **misdemeanor**: a single misdemeanor only results in a financial penalization of the node. Accumulating multiple misdemeanors might result in a node being subsequently ejected from the network.
     > Examples include:
     > • Execution or Collector nodes being involved in a Missing Collection Challenge
     > • Execution Receipt publishes an Execution Receipt that is unverified and conflicts with the sealed results (aka a potential Wintermute attack)

   To the best of my understanding, @jordanschalm's proposal is essentially using the flow of extension of handling `misdemeanors` for all cases. 

Technically, both a valid approaches. They just differ on the time scales, how long it takes to eject a misbehaving node. I think this is a product question.

flips/20210118-self-ejection.md
```diff
@@ -0,0 +1,208 @@
...
+then the identity table change is propagated to the consensus committee (and ultimately to the protocol state) via a service event.
+
+![Illustration of the self-ejection process. For simplicity, we denote denote the node as well as the node operator requesting self-ejection as _Alice_ ](20210118-self-ejection/Self-Ejection.png)
+
+1. To request self-ejection, the operator (denoted as _Alice_) submits the respective transaction 
+   (denoted as _T_) to the network (same process as any other conventional transaction).
+   For security purposes, we might want to require that the request is signed by the staking key of the node that is requesting its own ejection.
+2. The transaction will eventually be included in a block (denoted as _A_ in the illustration above).
+
+   1. When the Execution Nodes process transaction _T_, they emit a `Service Event`.
```
jordanschalm, line 63:
```suggestion
   1. When the Execution Nodes process transaction _T_, a `Service Event` is emitted. 
```

flips/20210118-self-ejection.md
```diff
@@ -0,0 +1,208 @@
...
+
+This matches the flow of information we have for identity table changes during epoch transitions:
+nodes submit staking requests to a smart contract, the requests are validated and aggregated into an identity table change,
+then the identity table change is propagated to the consensus committee (and ultimately to the protocol state) via a service event.
+
+![Illustration of the self-ejection process. For simplicity, we denote denote the node as well as the node operator requesting self-ejection as _Alice_ ](20210118-self-ejection/Self-Ejection.png)
+
+1. To request self-ejection, the operator (denoted as _Alice_) submits the respective transaction 
+   (denoted as _T_) to the network (same process as any other conventional transaction).
+   For security purposes, we might want to require that the request is signed by the staking key of the node that is requesting its own ejection.
```
jordanschalm, line 60: This transaction will require access to the resource obtained by the operator during staking registration. For additional security we may want to require that the request is signed by the staking key of the node requesting self-ejection.

jordanschalm (approved):

flips/20210118-self-ejection.md
AlexHentschel: much better wording. Included in https://github.com/onflow/flow/pull/324/commits/f4e178e46f2cdfbcc464a444dd686eadac71734f

PR #326, opened 2021-01-20
Update testnet-deployment.md

Dibri:
The parameter --code doesn't exist, for the path to the smart contract the correct parameter is --contract as in the example

## Description

I suppose that there is a typo in the doc, the parameter --code doesn't exist, and for the path to the smart contact I can use --contract


turbolent (approved):
Thank you for fixing this!

PR #386, opened 2021-02-18
Pack protobuf into a jar

m4ksio:
Adds gradle wrapper and build config so a simple JAR containing protobuf files can be built.
It will allow us to later distribute a JAR in appropriate repos and usage of JitPack for now 
For example - https://jitpack.io/com/github/onflow/flow/m4ksio~pack-protobuf-into-jar-0a22dbbe69-1/


turbolent (approved):
Nice! Great idea

psiemens (approved):
🎉

PR #471, opened 2021-04-12
Add an overview of all available SDKs

turbolent:
It would be nice to integrate this into the docs, maybe as a link in the left sidebar.

______

For contributor use:


githubwei:
@turbolent I was wondering if we need to add a disclaimer clarifying many SDKs are being maintained by the community, as volunteers, so before using developers should be aware there is no guarantee in terms of continued development and support. 

That way we avoid the issue of some developers complaining they wasted their efforts in building an app with the assumption of the SDK being always available and maintained, especially with breaking changes.

turbolent:
@githubwei This is true for any open-source project, so I'd let such considerations be up to the user

turbolent:
@psiemens @10thfloor any thoughts?

srinjoyc:
I'd add here as well that I'd agree with @githubwei  => We should point out which repos have complete documentation and is "recommended" by us. With so many choices, it might be easy to assume there is equal support from the community which is not quite true. Their experience could end quite quickly if they get frustrated and make the wrong choice here.

srinjoyc (approved):

10thfloor (approved):

Issue #494, opened 2021-05-04
[Analytics] Add Mixpanel to all docs pages


srinjoyc:
### Why?

Currently we are only tracking a limited number of events on the docs site, most of which are on the playground. We want to collect more data around the usage of the docs site to inform better design and product decisions. The goal of this first pass is to collect all critical points of interaction first and set up a standardized event format to use for the future.

### What? 

1. Each page (unique URL) should submit an event with the following:
     - Event Name: `${pageName}_viewed` where `pageName` is the url

 **Ensure that this is only run ONCE per page view**

Test the following cases:
https://docs.onflow.org/flow-go-sdk should map to an event name of `flow-go-sdk_viewed`
https://docs.onflow.org/flow-go-sdk/building-transactions/ should map to an event name of `flow-go-sdk/building-transactions_viewed`

2. On the 'Getting Started' homepage, put in events for each link (not on the navigation):
    - Event Name: `Homepage_link_${linkName}_clicked` where `linkName` is the url

3. On the left navbar on homepage, put in events for each nav link:
    - Event Name: `Homepage_nav_${linkName}_clicked` where `linkName` is the url

4. Put an event for the search bar (should be the same across any page).
    - Event Name: `Searched`
    - Properties: `{ text: userSearchTextString }`

5. Add an event on the 'Flow docs' logo that takes you back to the homepage. 
    - Event Name: `Home_clicked`

### Notes
- The existing Mixpanel API key should work for submitting these events.
- If possible, lets test out on a dev mixpanel account to ensure events are working as intended through some staging/local environment.

PR #503, opened 2021-05-18
Adds updated Mixpanel events

10thfloor:
Closes: #494 

## Description

Adds updated mixpanel events per #494


docs/content/components/tracking-link.js
```diff
@@ -0,0 +1,24 @@
...
+      }}
+    >
+      {children}
+    </a>
+  );
+};
+
+TrackingLink.propTypes = {
+  apiUrl: PropTypes.string,
+};
```
10thfloor, line 22:
```suggestion
TrackingLink.propTypes = {
  href: PropTypes.string,
  eventName: PropTypes.string
};
```

srinjoyc (approved):

PR #526, opened 2021-06-17
Network status isn't accurate, remove it.

10thfloor:
Original code is not removed, so we can re-use it once there is a more robust way of reporting network status.


docs/plugins/gatsby-theme-flow/src/components/page-layout.js
```diff
@@ -201,7 +201,7 @@ export default function PageLayout(props) {
               <MobileLogo />
               <MenuButton onClick={openSidebar} />
             </MobileNav>
-            <StatusWidget />
+            {/* <StatusWidget /> */}
```
10thfloor, line 204:
```suggestion
            {/* Leave the widget here, so we can re-enable when status endpoint is ready. */}
            {/* <StatusWidget /> */}
```

orodio (approved):

PR #538, opened 2021-06-21
Better landing page.

10thfloor:


docs/.env.development
```diff
@@ -0,0 +1,2 @@
+GATSBY_STATUSPAGE_API_KEY=e6278eb7-27f1-4795-adce-9a0751a4a31c
+GATSBY_STATUSPAGE_PAGE_ID=ytw5bdg6zr13
```
kimcodeashian, line 2: 🚩 Not sure if we want to expose this ?

kimcodeashian:
image

The colours for the footer is quite interesting 😛

kimcodeashian:
image

Is this too overwhelming? Maybe we can have a recommendation of where to start. It can be daunting sometimes w/ too many choices. Example: 
 
Don't know where to start?
Check out: Cadence & Flow Playground!

Issue #535, opened 2021-06-21
[Access API] upgrade to buf from prototool


smnzhu:
## Context

Currently, we use prototool to generate gRPC clients from the protobuf spec:

https://github.com/onflow/flow/tree/master/protobuf

However, it seems like this project is no longer actively maintained, and in particular the README on the repo itself recommends using a newer tool called Buf: https://github.com/uber/prototool

Another benefit of migrating is that it will be easier to configure `gRPC-gateway` to work properly, since it supports configuration in Buf: https://github.com/grpc-ecosystem/grpc-gateway#:~:text=here's%20an%20example%20buf.gen.yaml%20you%20can%20use%20to%20generate%20the%20stubs%20with%20buf%3A

We will likely be using `gRPC-gateway` to generate the reverse-proxy for the Access Node REST API which is currently a work in progress.

@psiemens @JeffreyDoyle @orodio: Vishal informed me that this may be a breaking change for the JS and Go SDKs?

## Definition of Done

What do we need to achieve to consider this issue resolved? For example:

* migrated to buf instead of prototool
* `make generate` works properly with buf

## PR Review 

Who is your Code Partner - @vishalchangrani 
Who can provide external eyes - @psiemens @JeffreyDoyle @orodio

## Internal Testing 
Estimation of core issue should capture larger testing efforts

## Estimation Breakdown
- Dev work: 2
- Testing: 1
- Review: 1

PR #546, opened 2021-06-23
migrate prototool to buf

smnzhu:
Closes: #???

## Description

Migrates prototool to buf, and updates the documentation.


buf.lock
```diff
@@ -0,0 +1,9 @@
+# Generated by buf. DO NOT EDIT.
```
vishalchangrani, line 1: do we need to check-in this `lock` file?

protobuf/README.md
```diff
@@ -8,12 +8,12 @@ Check out the [Flow Access API specification](/docs/content/access-api-spec.md).
 
 ## Generating Go stubs
 
-You can use [prototool](https://github.com/uber/prototool) to generate gRPC client stubs in a variety of languages.
+You can use [buf](https://github.com/bufbuild/buf) to generate gRPC client stubs in a variety of languages.
 
-Running the command below (in the current directory) will generate stubs for Go:
+Running the command below (in the root directory) will generate stubs for Go:
```
vishalchangrani, line 13: so the `flow` repo has a lot more than the protobuf files. It also has all the documentation under `content` and the node operator list under `nodeoperators`. Hence, it would be best if we can keep the `makefile` and other related things to generate the protobuf go code under the `protobuf` directory instead of outside.

protobuf/README.md
```diff
@@ -8,12 +8,12 @@ Check out the [Flow Access API specification](/docs/content/access-api-spec.md).
...
 
 ```shell script
-make generate
+make generate-go-stubs
```
vishalchangrani, line 16: Prerequisite: Install [buf](https://docs.buf.build/installation/)
```suggestion
make generate-go-stubs
```

protobuf/README.md
```diff
@@ -8,12 +8,12 @@ Check out the [Flow Access API specification](/docs/content/access-api-spec.md).
 
 ## Generating Go stubs
 
-You can use [prototool](https://github.com/uber/prototool) to generate gRPC client stubs in a variety of languages.
+You can use [buf](https://github.com/bufbuild/buf) to generate gRPC client stubs in a variety of languages.
 
-Running the command below (in the current directory) will generate stubs for Go:
+Running the command below (in the root directory) will generate stubs for Go:
```
vishalchangrani, line 14: Prerequisite: Install [buf](https://docs.buf.build/installation/)

buf.lock
smnzhu: Yes, it is supposed to be chedked in

protobuf/README.md
smnzhu: Yes, this is reflected in the README updates.

protobuf/README.md
smnzhu: Addressed comment in 5c12bdd

protobuf/go/flow/access/access.pb.go
```diff
@@ -1,2577 +1,1570 @@
...
-// versions:
-// 	protoc-gen-go v1.26.0
-// 	protoc        v3.17.1
 // source: flow/access/access.proto
 
 package access
 
 import (
+	fmt "fmt"
+	proto "github.com/golang/protobuf/proto"
```
vishalchangrani, line 8: so `buf` seems to be reverting to the older protobuf implementation for golang - 

This module (github.com/golang/protobuf) contains Go bindings for protocol buffers.

see [here](https://github.com/golang/protobuf#go-support-for-protocol-buffers) - 

> It has been superseded by the google.golang.org/protobuf module, which contains an updated and simplified API, support for protobuf reflection, and many other improvements. We recommend that new code use the google.golang.org/protobuf module.

vishalchangrani: nevermind..we are currently using the old `github.com/golang/protobuf/proto` as well.

protobuf/flow/access/access.proto
```diff
@@ -17,78 +17,101 @@ import "google/protobuf/timestamp.proto";
 service AccessAPI {
 
   // Ping is used to check if the access node is alive and healthy.
-  rpc Ping (PingRequest) returns (PingResponse);
+  rpc Ping(PingRequest) returns (PingResponse);
```
vishalchangrani, line 20: can you reset the indentation to what it was. This is the plugin I use to set the indentation automatically -  https://plugins.jetbrains.com/plugin/8277-protobuf-support

vishalchangrani (approved):
lgtm

Kay-Zee:
Could we either include more detail on reasoning for this change in the PR, or at least link to an issue that does so?

smnzhu:
> Could we either include more detail on reasoning for this change in the PR, or at least link to an issue that does so?

Hey Kan, there is an issue already attached, is it not showing up?

smnzhu:
fixes #535

PR #559, opened 2021-07-07
Epoch Docs Organization Proposal

joshuahannan:
Closes https://github.com/onflow/flow-core-contracts/issues/209

## Description

I'm targeting this to the `feature/epochs` branch, so it is meant to be a organizational proposal. The empty sections and those marked as `TODO` will be filled in in other PRs to this feature branch.

* Adds a docs structure for the inclusion of Epoch and staking collection docs.
* I've basically split up the technical docs from the actual user guides
* Updates the basic overview to include some information about epochs.
* Removes unnecessary docs and renames/combines others
* I combined the staking and delegating docs for unlocked and locked flow because we are effectively deprecating those methods and highly encouraging people to use the staking collection instead, but we shouldn't remove them completely in case people stay with the old version. It made sense to consolidate the pages so we have less in the sidebar.

______

For contributor use:


docs/content/staking/unlocked-staking-guide.mdx
```diff
@@ -1,10 +1,13 @@
...
+title: Basic Staking with Unlocked FLOW
+sidebar_title: Basic Staking
 ---
 
 This document outlines the steps a token holder can take to stake and manage 
-a Flow node with normal, unlocked FLOW using the `FlowIDTableStaking` contract.
+a Flow node with normal, unlocked FLOW using only the types defined in the `FlowIDTableStaking` contract.
+It only supports having one node or delegator object per account and is not supported by ledger
+and will likely not be supported by other wallets, so it is recommended to use the staking collection
+instead. This style of staking is for those who want to keep their staking operation as simple as possible.
```
jordanschalm, line 10:
```suggestion
instead.
```

Sounded like we're deprecating this staking method - not sure we want to give people a reason to keep using it if that's the case.

jordanschalm (approved):
I really like the separation of "user guide" and "technical overview" 👏. Makes a lot of sense.

I only reviewed the epoch-related changes, but the organization and updated docs for epochs looks great to me. Thanks!

PR #560, opened 2021-07-08
Execution result distribution

m4ksio:
Adds new method with entities to enable retrieval of execution results.
As this is a new method and new entities and nothing else was modified this shouldn't cause any compatibility issues, simply adding new endpoint.
This is required to expose execution results which contains computed events hashes which would enable ability for clients to verify the events emitted.


smnzhu:
Can you explain how exactly this is different from the existing `GetTransactionResult`?

smnzhu (approved):

m4ksio:
> Can you explain how exactly this is different from the existing `GetTransactionResult`?

Transaction results gives you, well, single transaction result. This one gives you execution result for entire block, which also contains chunk data which include events hash.

protobuf/flow/entities/execution_result.proto
```diff
@@ -0,0 +1,29 @@
...
+
+message ExecutionResult {
+  bytes previous_result_id = 1;
+  bytes block_id = 2;
+  repeated Chunk chunks = 3;
+  repeated ServiceEvent service_events = 4;
+}
+
+message Chunk {
+  uint32 CollectionIndex = 1;
```
Kay-Zee, line 16:
```suggestion
  uint32 collection_index = 1; // index of collection within the block associated with this chunk
```

protobuf/flow/entities/execution_result.proto
```diff
@@ -0,0 +1,29 @@
...
+  bytes start_state = 2; // state when starting executing this chunk
+  bytes event_collection = 3; // Events generated by executing results
+  bytes block_id = 4; // Block id of the execution result this chunk belongs to
+  uint64 total_computation_used = 5; // total amount of computation used by running all txs in this chunk
+  uint32 number_of_transactions = 6; // number of transactions inside the collection
```
Kay-Zee, line 21: i noticed this went from uint64 in flow-go to uint32 here. Is the thought that this just won't need that much number space?

I feel it might still be better to match types, even if it's excessive

protobuf/go.mod
```diff
@@ -4,5 +4,5 @@ go 1.16
 
 require (
 	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.1.0
-	google.golang.org/protobuf v1.26.0
+	google.golang.org/protobuf v1.5.0
```
Kay-Zee, line 7: did we downgrade? what's the reason for this here?

Kay-Zee (commented):
Question about the protobuf version, and a few minor comments

protobuf/flow/entities/execution_result.proto
smnzhu: Yeah, actually I noticed this isn't the only case where this is true either. There are a few other cases where the field doesn't match that in the Go model

m4ksio: > Is the thought that this just won't need that much number space?

Yes, I guess we always said 1 000 000 is max number of tx per block.
I'm not sure if its worth expanding to uint64 here or simply reducing to uint32 in main flow repo

protobuf/go.mod
m4ksio: flow-go-sdk uses even older `1.4.x` and there were compatibility and compilation issues. I will check maybe if we can bump `flow-go-sdk` instead.

vishalchangrani:
@m4ksio - if you don't mind, can you also update the documentation - https://docs.onflow.org/access-api/ (which comes from - https://github.com/onflow/flow/blob/master/docs/content/access-api.md) thanks!

protobuf/flow/entities/execution_result.proto
```diff
@@ -0,0 +1,29 @@
...
+  bytes previous_result_id = 1;
+  bytes block_id = 2;
+  repeated Chunk chunks = 3;
+  repeated ServiceEvent service_events = 4;
+}
+
+message Chunk {
+  uint32 CollectionIndex = 1;
+  bytes start_state = 2; // state when starting executing this chunk
+  bytes event_collection = 3; // Events generated by executing results
```
vishalchangrani, line 18: are these different from these events https://docs.onflow.org/access-api/#event?

m4ksio: No, this is hash of the events emitted

protobuf/flow/entities/execution_result.proto
m4ksio: Reverted to uint64 to match whats in 'flow-go'

m4ksio:
> @m4ksio - if you don't mind, can you also update the documentation - https://docs.onflow.org/access-api/ (which comes from - https://github.com/onflow/flow/blob/master/docs/content/access-api.md) thanks!

Great suggestion! Added, would appreciate your review on it

docs/content/access-api.md
```diff
@@ -790,9 +790,44 @@ message ProtocolStateSnapshotResponse {
...
 ```
+</details>
+
+## Execution results
+
+The following method can be used to query the for [execution results](https://github.com/onflow/flow-go/blob/master/model/flow/execution_result.go) for a given block.
+
+### GetExecutionResultForBlockID
+
+`GetExecutionResultForBlockID` retrieves execution result for given block. It is different from Transaction Results,
```
vishalchangrani, line 801: ❤️

vishalchangrani (approved):
lgtm

protobuf/flow/entities/execution_result.proto
vishalchangrani: got it

PR #562, opened 2021-07-12
Add a line on the difference between sepc256k1 & secp256r1.

huitseeker:
Clarifies ECDSA algorithms

### DIFF
--- a/docs/content/concepts/accounts-and-keys.md
+++ b/docs/content/concepts/accounts-and-keys.md
@@ -82,7 +82,7 @@ Flow will have initial support for a predefined set of signature and hash pairin
 | ECDSA     | P-256     | ECDSA_P256      | 2    |
 | ECDSA     | secp256k1 | ECDSA_secp256k1 | 3    |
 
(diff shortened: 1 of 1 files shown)


docs/content/concepts/accounts-and-keys.md
```diff
@@ -82,6 +82,9 @@ Flow will have initial support for a predefined set of signature and hash pairin
 | ECDSA     | P-256     | ECDSA_P256      | 2    |
 | ECDSA     | secp256k1 | ECDSA_secp256k1 | 3    |
 
+There are two curves commonly used with the ECDSA algorithm, secp256r1 ([OID 1.2.840.10045.3.1.7](http://oid-info.com/get/1.2.840.10045.3.1.7), also called the "NIST P-256." curve), and secp256K1 ([OID 1.3.132.0.10](http://oid-info.com/get/1.3.132.0.10), the "Bitcoin" curve). Please be sure to double-check which parameters you are using before registering a key, as presenting a key using one of the curves under the code and format of the other will generate an error.
```
tarakby, line 85: - capital `K` to `k` in the curve name.
- a slight wording change to avoid defining the curve as the bitcoin curve, but as an existing curve that happened to be used by bitcoin. 

```suggestion
There are two curves commonly used with the ECDSA algorithm, secp256r1 ([OID 1.2.840.10045.3.1.7](http://oid-info.com/get/1.2.840.10045.3.1.7), also called the "NIST P-256." curve), and secp256k1 ([OID 1.3.132.0.10](http://oid-info.com/get/1.3.132.0.10), the curve used by "Bitcoin"). Please be sure to double-check which parameters you are using before registering a key, as presenting a key using one of the curves under the code and format of the other will generate an error.
```

tarakby (approved):
Thanks for updating the doc 🏄 

I am not sure if specifying the OIDs is necessary in the doc, since libraries usually refer to curves by their names rather than OIDs, but it's fine if we keep it.

PR #567, opened 2021-07-14
Epoch Docs

joshuahannan:
Closes: https://github.com/onflow/flow-core-contracts/issues/210

## Description

Adds docs for the Epoch preparation protocol, `FlowEpoch` smart contract, service events, epoch metadata, and epoch scripts.

______

For contributor use:


docs/content/staking/epoch-preparation.mdx
```diff
@@ -16,28 +16,143 @@ description: Technical Overview of the Flow Epoch Protocol
...
+  **Motivation for measuring epoch length in consensus views:**
+
+
+  While we utilize a consensus-internal detail for measuring time,
+  this solution has significantly less engineering complexity compared to using height.
+  The reason is that HotStuff requires (for liveness) a deterministic *fork-independent* primary selection per view.
+  Defining Epoch length based on height would break fork-independent primary selection,
+  as there can be multiple blocks at the same height, but with different views.
+  Using view numbers, there are no extra edge cases. 
+</Callout>
```
jordanschalm, line 108: I feel like this might be too much detail for these docs. To me the important thing to be aware of is that epochs are measured in views, not heights, and that epoch length is defined before the epoch starts.

jordanschalm (commented):
Thanks Josh, this is looking great.

Added some thoughts from initial pass.

docs/content/staking/epoch-preparation.mdx
joshuahannan: That is a good point. I'll trim this dowm

PR #582, opened 2021-07-21
Document time synchronization in running a Flow node

huitseeker:
Helps with NTP setup at the host.

### DIFF
--- a/docs/content/node-operation/node-setup.mdx
+++ b/docs/content/node-operation/node-setup.mdx
@@ -72,6 +72,20 @@ Flow also provides `systemd` service and unit files as a template for installati
   those are good choices.
 </Callout>
 
+## Time synchronization
+
+You should also ensure you run **time synchronization** on the machine hosting the container, to avoid clock drift. In practice, this means configuring a client for the NTP protocol, and making sure it runs as a daemon. `ntpd` is one recommended example. To configure it, you just have to point it to an NTP server to query periodically. A default from your Linux distribution or cloud operator may already be set, and in the interest of decentralization, our recommendation would be to use it unless you have a specific reason to do otherwise.
+
+<Callout type="info" title="Time synchronization FAQ">
+
+- **Leap-smearing**: Leap-smearing time servers and non-leap-smearing time servers are both acceptable for the magnitude of our time precision requirements - though considering very few providers offer leap smearing time servers, a "regular" time server helps ensure our pool of time providers is more diverse.
+
+- **Why not do it in the container itself? Why do we need to do this?**:  Without special privileges and in all major container runtimes, a container will not run with the `CAP_SYS_TIME` capability. For Flow, this means that the node software itself cannot change the time of the host machine, making the in-container use of standard time synchronization protocols ineffective.
+
+- **Why does time matter in Flow?**: Time information comes up in consensus and in smart contracts. The consensus algorithm of Flow allows nodes to exit the influence of a corrupt or ineffective "leader" node by collectively deciding to switch to the next "phase" of the protocol at the right time. The smart contract language also allows developer access to block time stamps, which provide an approximation of time. To resist manipulation in each case, honest nodes must compute timing values from an aggregate of the information provided by all nodes. That approach, though resilient, is still sensitive to inaccurate time information. In other words, a node subject to clock drift but otherwise honest will not stop the consensus, but might make it slower.
+
+</Callout>
+
 ## Setup Data Directories & Disks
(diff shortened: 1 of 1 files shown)


docs/content/node-operation/node-setup.mdx
```diff
@@ -72,6 +72,18 @@ Flow also provides `systemd` service and unit files as a template for installati
   those are good choices.
 </Callout>
 
+You should also ensure you run **time synchronization** on the machine hosting the container, to avoid clock drift. In practice, this means configuring a client for the NTP protocol, and making sure it runs as a daemon. `ntpd` is one recommended example. To configure it, you just have to point it to an NTP server to query periodically. A default from your Linux distribution or cloud operator may already be set, and in the interest of decentralization, our recommendation would be to use it unless you have a specific reason to do otherwise.
+
+<Callout type="info" title="TIme synchronization FAQ">
```
vishalchangrani, line 77:
```suggestion
<Callout type="info" title="Time synchronization FAQ">
```

docs/content/node-operation/node-setup.mdx
```diff
@@ -72,6 +72,18 @@ Flow also provides `systemd` service and unit files as a template for installati
   those are good choices.
 </Callout>
 
+You should also ensure you run **time synchronization** on the machine hosting the container, to avoid clock drift. In practice, this means configuring a client for the NTP protocol, and making sure it runs as a daemon. `ntpd` is one recommended example. To configure it, you just have to point it to an NTP server to query periodically. A default from your Linux distribution or cloud operator may already be set, and in the interest of decentralization, our recommendation would be to use it unless you have a specific reason to do otherwise.
```
vishalchangrani, line 75:
```suggestion

You should also ensure you run **time synchronization** on the machine hosting the container, to avoid clock drift. In practice, this means configuring a client for the NTP protocol, and making sure it runs as a daemon. `ntpd` is one recommended example. To configure it, you just have to point it to an NTP server to query periodically. A default from your Linux distribution or cloud operator may already be set, and in the interest of decentralization, our recommendation would be to use it unless you have a specific reason to do otherwise.
```

vishalchangrani: extra new line

docs/content/node-operation/node-setup.mdx
```diff
@@ -72,6 +72,18 @@ Flow also provides `systemd` service and unit files as a template for installati
   those are good choices.
 </Callout>
```
AlexHentschel, line 74: Shall we add a newline here? Git renders this without any section separation (screenshot): 
Screen Shot 2021-07-26 at 10 31 17 AM

```suggestion
### Time synchronization

```

AlexHentschel (approved):
👍

vishalchangrani (approved):

docs/content/node-operation/node-setup.mdx
```diff
@@ -72,6 +72,20 @@ Flow also provides `systemd` service and unit files as a template for installati
...
+
+- **Why not do it in the container itself? Why do we need to do this?**:  Without special privileges and in all major container runtimes, a container will not run with the `CAP_SYS_TIME` capability. For Flow, this means that the node software itself cannot change the time of the host machine, making the in-container use of standard time synchronization protocols ineffective.
+
+- **Why does time matter in Flow?**: TIme information comes up in consensus and in smart contracts. The consensus algorithm of Flow allows node to exit the influence of a corrupt or ineffective "leader" node by collectively deciding to switch to the next "phase" of the protocol at the right time. The smart contract language also allows developer access to block time stamps, which provide an approximation of time. To resist manipulation in each case, honest nodes must compute timing values from an aggregate of the information provided by all nodes. That approach, though resilient, is still sensitive to inaccurate time information. In other words, a node subject to clock drift but otherwise honest will not stop the consensus, but might make it slower.
```
smnzhu, line 85:
```suggestion
- **Why does time matter in Flow?**: Time information comes up in consensus and in smart contracts. The consensus algorithm of Flow allows nodes to exit the influence of a corrupt or ineffective "leader" node by collectively deciding to switch to the next "phase" of the protocol at the right time. The smart contract language also allows developer access to block time stamps, which provide an approximation of time. To resist manipulation in each case, honest nodes must compute timing values from an aggregate of the information provided by all nodes. That approach, though resilient, is still sensitive to inaccurate time information. In other words, a node subject to clock drift but otherwise honest will not stop the consensus, but might make it slower.
```

PR #585, opened 2021-07-27
Update MD5 -> SHA256

huitseeker:
Do not merge before https://github.com/onflow/flow-go/pull/990/ is merged.
This PR updates our file-based user-visible hashes from MD5 to SHA256, a better, yet widely available hash function.

The present PR does the same for the doc, at least in the format of the displayed hashes, rather than actual content: the doc is not updated at every Spork, and the output updated in this PR is here for demonstration only.


AlexHentschel (approved):
🌻

vishalchangrani (commented):
sry for the delay

PR #601, opened 2021-08-17
Update site to use Gatsby v3

10thfloor:
Faster build times. Updated React -> MDX pipeline.

All project dependencies were updated to latest compatible versions. Using the process described here: 
https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v2-to-v3


docs/development/cadence-downstream-dependencies.md
```diff
@@ -1,38 +0,0 @@
-# Updating the Cadence Downstream Dependencies / Releasing the CLI
```
psiemens, line 1: Can we keep this file? It's still helpful as a reference.

psiemens (changes requested):

docs/development/cadence-downstream-dependencies.md
10thfloor: Yes, I'll PR this back in.

PR #631, opened 2021-09-14
FLIP: Voting Mechanics in Decentralized Ecosystems

JeffreyDoyle:
FLIP: Voting Mechanics in Decentralized Ecosystems

### DIFF
--- a/flips/20210914-voting-mechanics-decentralized-ecosystems.md
+++ b/flips/20210914-voting-mechanics-decentralized-ecosystems.md
@@ -2,10 +2,10 @@
 
 | Status        | Proposed                                             |
 :-------------- |:---------------------------------------------------- |
-| **FLIP #**    |                                                      |
+| **FLIP #**    | [631](https://github.com/onflow/flow/pull/631)       |
 | **Author(s)** | Jeffrey Doyle (jeffrey.doyle@dapperlabs.com),        |
 | **Sponsor**   | Jeffrey Doyle (jeffrey.doyle@dapperlabs.com)         |
-| **Updated**   | 2020-09-14                                           |
+| **Updated**   | 2021-09-14                                           |
 
 ## Overview
 
--- a/flips/20210914-voting-mechanics-decentralized-ecosystems.md
+++ b/flips/20210914-voting-mechanics-decentralized-ecosystems.md
@@ -1,0 +1,119 @@
+# Voting Mechanics in Decentralized Ecosystems
+
+| Status        | Proposed                                             |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    |                                                      |
+| **Author(s)** | Jeffrey Doyle (jeffrey.doyle@dapperlabs.com),        |
+| **Sponsor**   | Jeffrey Doyle (jeffrey.doyle@dapperlabs.com)         |
+| **Updated**   | 2020-09-14                                           |
+
+## Overview
+
+In decentralized organizations, governance is often constructed such that
+authority in the organization is distributed horizontally to many or all
+organizational participants, rather than vertically in such a way that grants
+some participants authority in decision making while others do not have a voice.
+
+The topic of decentralized governance has been explored many times across
+various blockchain projects. Often a DAO (Decentralized Autonomous Organization)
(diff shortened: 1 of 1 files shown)


flips/20210914-voting-mechanics-decentralized-ecosystems.md
```diff
@@ -0,0 +1,119 @@
...
+ability to impact change upon that system. The actions performed by the DAO upon
+the system it controls are often _proposed_, and then _voted_ upon by members or
+participants of the organization. If conditions are met such that a vote is
+passed, the proposed action is carried out autonomously by the DAO.
+
+What a DAO can do is virtually limitless. However, the mechanic that enables
+community participation in the actions performed by a DAO, or any other
+decentralized ecosystem is contentious. There has yet to be discovered a
+universally accepted set of best practices for enabling decentralized, community
+driven governance in a decentralized system.
```
janezpodhostnik, line 31:
```suggestion
decentralized ecosystem is contentious. A universally accepted set of best practices
for enabling decentralized, community driven governance in a decentralized system
is yet to be discovered.
```

Perhaps it sounds better this way around? 
Maybe a different verb could be used instead of _discovered_? _crafted_?

janezpodhostnik (commented):
Great writeup ❤️! I guess this is more of a debate starter for something that will eventually become a proposal. Will the debate take place here in the comments?

JeffreyDoyle:
@janezpodhostnik Debate is encouraged to take place in this issues comments!

janezpodhostnik:
# An opinion on counting votes

If we consider it a fact that there will always be some that want to skew the vote to their side by any means necessary, wouldn't it be interesting to give them a way to do so?

If by every act of voting the community/system would benefit or grow, then by casting more votes to
try to sway the results in your favour you would also be improving the community as a whole.

An example could be that for every vote case you must personally plant a tree, fix a road,
or help an old lady cross the street. A voting *battle* could leave the
neighbourhood in much better state in the end. We just have to be careful
that the system doesn't run out of old ladies.

Care should also be taken to not tie the ability to cast more votes (or increase vote weight)
to the wealth of a person. If saplings are expensive then those
with more funds will have more opportunities to increase the weight of their vote.

## How does this tie into Flow?

There should be a neutral fund that benefits everyone in a way, and there is one: Transaction Fees.
Using the system is good for the system. 
Collecting a lot of transaction fees will cause deflation which is also good for the system.
The idea would then be that the weight of an accounts' vote would be 
(linearly!) proportional to the amount of transactions fees paid by that account.
Accounts that have more funds would have more opportunity to influence the vote, but they 
still couldn't spam the system with transactions, as the transaction fees would increase because of it. Another measure that could also be put in place is to increase the fees of a single account 
if there is a single account spamming our system... 
The interesting part here is that these mitigations are something that is needed to prevent spam anyway.

This raises the question about custodies that pay fees for their accounts. These custodies would have a lot
of vote weight, while their users might have none. I'm my opinion, if the users of these custodies
are aware of this fact, I think this is not a problem at all.

## An unrelated note about threshold conditions

I think any threshold conditions for letting people vote are not the way to go.

*What?*

By threshold conditions I mean something like the following:

- only accounts over x FLOW have the ability to vote
- only accounts over x years old have the ability to vote
- only accounts that _insert condition_ have the ability to vote

*Why?*

It is relatively cheap to create new accounts. people that have the means can create many accounts and make sure they individually pass the threshold to be able to vote, while people that don't have the means wont be able to do this.

bluesign:
> An example could be that for every vote case you must personally plant a tree, fix a road, or help an old lady cross the street. 

@janezpodhostnik need to be very careful here, people can start to "break" roads to "fix" them. There is a very nice story with snakes about this. 

For the voting part, I think it all depends on the power of voting. I think the document should give some examples, for me, it is the first time I am reading about the subject, and little confusing without some examples. Will it be direct democracy or representative? 

The voting problem is also very hard to fix, you need to balance it with something else,  if I am voting 1 per account, having more flow in 1 account has to be better than having the same flow in N accounts. 

Also, there is the issue of governance, let's say we voted for resolution 1, how long it stays in effect, can we challenge it the next day with opposite resolution number 2?

PR #634, opened 2021-09-16
Adds local copy of SDK guidelines.

10thfloor:
The content is cloned from the guidelines repo, but must exist as part of this repo as well, so. we can customize the docs site menus.


10thfloor:
This is no longer necessary. SDK guidelines are pulled from respective Github repo.

PR #636, opened 2021-09-17
FLIP-0636: NFT Metadata

briandilley:
<p align="center">
  
</p>

Closes: [flip-fest #16](https://github.com/onflow/flip-fest/issues/16)

## Description
Team _**[Non Fungibles](https://www.hackerearth.com/challenges/hackathon/flip-fest/dashboard/3744a70/team/)**_ here submitting our proposal to the NFT Meta Data spec for the flip-fest.

Authors include:
- @bjartek 
- @bluesign
- @briandilley  

______

For contributor use:

### DIFF
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -17,7 +17,7 @@ work for any contract and any resource, not just a single one.
 ## Motivation
 Currently there is no standard for defining NFT/resource metadata. At best, most implementations 
 employ a simple `{String: AnyStruct}` dictionary. Due to the various methods that NFT 
-authors use it’s hard / impossible to create applications that use NFTs whether it be 
+authors use, it’s hard / impossible to create applications that use NFTs whether it be 
 for something as simple as viewing them or as complex as integrating them into a game 
 or application.
 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -104,7 +104,7 @@ pub struct interface ViewConverter {
 
 The most basic example is having `String` be a view. Think of this as `toString()` in Java or a `Stringer` implementation in Go. It is a way to view your NFT as a string. 
 
-The bigger brother of `String` would be `Display`. How do you preview an NFT in blocot, a marketplace or other solutions:
+The bigger brother of `String` would be `Display`. Here is how you could display an NFT in a wallet, marketplace or other user-facing application:
 ```
 	pub struct Display{
 		pub let name: String
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -251,28 +251,6 @@ pub struct URLWithHash {
 }
 ```
 
-### Royalty 
-
-This particular problem has been discussed extensively. The following suggestion was agreed upon
-by @bjartek, @rheaplex and @dete. 
-
-```
-  /// a struct interface that describes how to calculate royalty 
-	pub struct interface Royalty {
-
-		/// if nil cannot pay this type
-		/// if not nil withdraw that from main vault and put it into distributeRoyalty 
-		pub fun calculateRoyalty(type:Type, amount:UFix64) : UFix64?
-
-		/// call this with a vault containing the amount given in, calculate royalty and it will be distributed accordingly
-		pub fun distributeRoyalty(vault: @FungibleToken.Vault) 
-
-		/// generate a string that represents all the royalties this NFT has for display purposes
-		pub fun displayRoyalty() : String?  
-
-	}
-```	
-
 ## Drawbacks
 This standard does not include any required ways to store the metadata, and therefore some of the metadata 
 could be dynamic, depending on how the NFTs are authored. This is a double edged sword; it allows for greater 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -190,7 +190,7 @@ struct Artwork {
     let artistName: String
 }
 
-...
+// ...
 
 let artwork = nft.resolveView(Type<ArtNFT.Artwork>()) as? ArtNFT.Artwork
 ```
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -2,10 +2,10 @@
 
 | Status        | Proposed |
 :-------------- |:---------------------------------------------------- |
-| **FLIP #**    | [636](https://github.com/onflow/flow/pull/636)                                                                                     |
-| **Author(s)** | Bjarte Karslen (bjartek@find.xyz), Deniz Mert Edincik (you@example.org), Brian Dilley (briandilley@briandilley.com) |
-| **Sponsor**   | Peter Siemens (peter@dapperlabs.com)                                                                                               |
-| **Updated**   | 2021-12-06                                                                                                                         |
+| **FLIP #**    | [636](https://github.com/onflow/flow/pull/636)                                                                                       |
+| **Author(s)** | Bjarte Karslen (bjartek@find.xyz), Deniz Mert Edincik (Deniz@edincik.com), Brian Dilley (briandilley@briandilley.com) |
+| **Sponsor**   | Peter Siemens (peter@dapperlabs.com)                                                                                                 |
+| **Updated**   | 2021-12-06                                                                                                                           |
 
 
 ## Objective
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -288,54 +288,6 @@ by @bjartek, @rheaplex and @dete.
 	}
 ```	
 
-### Versus
-
-The versus project at versus.auction can be adapted to the new standard using these few lines of code
-
-```
-
-	 pub fun getViews() : [Type] {
-			return [
-			Type<String>(),
-			Type<TypedMetadata.Display>(),
-			Type<TypedMetadata.Editioned>(),
-			Type<TypedMetadata.CreativeWork>(),
-			Type<{TypedMetadata.Royalty}>(),
-			Type<TypedMetadata.Media>()
-			]
-		}
-		
-	 pub fun resolveView(_ type: Type): AnyStruct {
-
-			if type== Type<String>() {
-				return self.name.concat(" ").concat(self.metadata.edition.toString()).concat(" of ").concat(self.metadata.maxEdition.toString()).concat(" by ").concat(self.metadata.artist)
-			}
-
-			if type == Type<TypedMetadata.Display>() {
-				let description=self.metadata.description.concat(" by:").concat(self.metadata.artist).concat( " edition ").concat(self.metadata.edition.toString()).concat( " of ").concat(self.metadata.maxEdition.toString())
-				return TypedMetadata.Display(name: self.metadata.name, thumbnail: self.url!, description: description, source: "versus")
-			}
-
-			if type == Type<TypedMetadata.Editioned>() {
-				return TypedMetadata.Editioned(edition: self.metadata.edition, maxEdition:self.metadata.maxEdition)
-			}
-			if type == Type<TypedMetadata.CreativeWork>() {
-				return TypedMetadata.CreativeWork(artist:self.metadata.artist, name: self.metadata.name, description: self.metadata.description, type:self.metadata.type)
-			}
-			if type == Type<{TypedMetadata.Royalty}>() {
-				return Royalties(self.royalty)
-			}
-
-			if type == Type<TypedMetadata.Media>() {
-				if self.url != nil {
-					return TypedMetadata.Media(data: self.url!, contentType: self.metadata.type, protocol: "http")
-				}
-			}
-			return nil
-		}
-		
-```
-
 ## Drawbacks
 This standard does not include any required ways to store the metadata, and therefore some of the metadata 
 could be dynamic depending on how the NFTs are authored. This is a double edged sword; it allows for greater 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -71,7 +71,7 @@ This proposal by no means precludes the use of member variable on the NFT resour
 ideally all immutable data for an NFT is stored as immutable members of the NFT resource and views 
 are resolved and created by accessing those immutable members.
 
-### Supporting other resources besides NFT
+### Supporting other resources besides NFTs
 In order to support any other resource and resource Collection like marketplaces and FT the following new interface is proposed:
 ```
 pub resource interface ViewResolverCollection {
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -128,8 +128,8 @@ The data above is taken from what the [Blocto wallet](https://blocto.portto.io/)
 
 ### Other examples
 
-[For](For) example, an NFT implementation for artwork may want to allow for the image data to be retrieved 
-as an `HostedImage` struct:
+For example, an NFT artwork implementation may want to allow for image data to be retrieved 
+as an `IPFSImage` struct:
 ```
 struct IPFSImage {
     let url: String
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -21,7 +21,7 @@ authors use, it’s hard / impossible to create applications that use NFTs wheth
 for something as simple as viewing them or as complex as integrating them into a game 
 or application.
 
-If you want to create an Auction contract on Flow today you have to hardcode all the NFTs 
+If you want to create an Auction contract on Flow today, you have to hardcode all the NFTs 
 you want to support. This will often lead to duplicated contracts and code and just lots 
 of things to maintain. This standard will enable you to create a single Auction contract 
 and define the Type of views you need to expose in your NFT in order to support the contract. 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -2,10 +2,10 @@
 
 | Status        | Proposed |
 :-------------- |:---------------------------------------------------- |
-| **FLIP #**    | [636](https://github.com/onflow/flow/pull/636)                                                                         |
-| **Author(s)** | Bjarte Karslen (bjartek@find.xyz), Deniz Mert Edincik (you@example.org), Brian Dilley (you@example.org) |
-| **Sponsor**   | Peter Siemens (peter@dapperlabs.com)                                                                                   |
-| **Updated**   | 2021-12-06                                                                                                             |
+| **FLIP #**    | [636](https://github.com/onflow/flow/pull/636)                                                                                     |
+| **Author(s)** | Bjarte Karslen (bjartek@find.xyz), Deniz Mert Edincik (you@example.org), Brian Dilley (briandilley@briandilley.com) |
+| **Sponsor**   | Peter Siemens (peter@dapperlabs.com)                                                                                               |
+| **Updated**   | 2021-12-06                                                                                                                         |
 
 
 ## Objective
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -56,7 +56,7 @@ pub resource interface ViewResolver {
 ```
 
 We will also add default implementations for these methods so that no compatibility will be broken 
-for existing NFT solutions. The new NFT interface could look something like this: 
+for existing NFT solutions. The new NFT resource could look something like this: 
 ```
 pub resource NFT: INFT, ViewResolver {
     pub let id: UInt64
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -86,7 +86,7 @@ For single resources that are stored as is like a fungible token or a profile co
 
 ### Converting between different views
 
-Standards will evolve and new views will be added. In order to make it easier for maintainers we propose a mechanism to regsiter a converter from one View to another. 
+Standards will evolve and new views will be added. In order to make it easier for maintainers we propose a mechanism to register a converter from one View to another:
 
 ```
 pub struct interface ViewConverter {
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -168,7 +168,7 @@ data: QmRAQB6YaCyidP37UdDnjFY5vQuiBrcqdyoW1CuDgwxkD4
 contentType: image/jpeg
 protocol: ipfs
 
-An http image
+An HTTP image:
 data: https://test.find.xyz/find.png
 contentType: image/jpeg
 protocol: http
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -84,7 +84,7 @@ This will allow any collection to be exposed as view resolvable.
 
 For single resources that are stored as is like a fungible token or a profile contract you can just implement the `ViewResolver` interface directly.
 
-### Converting between different views. 
+### Converting between different views
 
 Standards will evolve and new views will be added. In order to make it easier for maintainers we propose a mechanism to regsiter a converter from one View to another. 
 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -39,7 +39,7 @@ will lead to more applications that integrate NFTs. This will make NFTs intrinsi
 useful and valuable to the end users holding the NFTs.
 
 A user will be much more likely to transfer their NFT to other marketplaces and sell their NFT 
-in different ways with this solution. Without it all marketplaces need to hardcode support 
+in different ways with this solution. Without it, all marketplaces need to hardcode support 
 for all NFTs directly. 
 
 It will also open up the possibility to standardize on some shared structs like Royalty 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -82,7 +82,7 @@ pub resource interface ViewResolverCollection {
 
 This will allow any collection to be exposed as view resolvable.
 
-For single resources that are stored as is like a FungibleToken or a Profile contract you can just implement the ViewResolver interface directly.
+For single resources that are stored as is like a fungible token or a profile contract you can just implement the `ViewResolver` interface directly.
 
 ### Converting between different views. 
 
--- a/flips/20210916-nft-metadata.md
+++ b/flips/20210916-nft-metadata.md
@@ -300,8 +300,8 @@ contract after it has already been published through the use of a `ViewResolver`
 [Playground available here](https://play.onflow.org/a71203b9-a03a-47f9-ab70-7285fcf4c56a?type=account&id=0)
 
 ## Compatibility
-Provided support for default methods on interfaces is added to cadence there are no backwards compatibility issues. 
-  Short of that, NFT authors will need to add implementations for the new methods added to their contracts.
+Provided support for default methods on interfaces is added to Cadence there are no backwards compatibility issues. 
+Short of that, NFT authors will need to add implementations for the new methods added to their contracts.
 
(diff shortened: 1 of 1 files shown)


flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,320 @@
...
+you want to support. This will often lead to duplicated contracts and code and just lots 
+of things to maintain. This standard will enable you to create a single Auction contract 
+and define the Type of views you need to expose in your NFT in order to support the contract. 
+So it will enable a design-by-contract pattern where the creator of a contract can define 
+the views it requires and an NFT can tell what views they export. 
+
+All NFTs and contracts using NFTs on flow are affected by this problem. This is most 
+visible in the number of contracts that Viv3 needs in order to support all their different 
+sellable items. With all the oncoming marketplaces/auction houses and things like that 
+coming to flow a good standard for NFTs will help everybody.
```
turbolent, line 26:
```suggestion
All NFTs and contracts using NFTs on Flow are affected by this problem. This is most 
visible in the number of contracts that Viv3 needs in order to support all their different 
sellable items. With all the oncoming marketplaces/auction houses and things like that 
coming to Flow a good standard for NFTs will help everybody. 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,320 @@
...
+    pub let id: UInt64
+    pub fun availableViews(): [Type] {
+       return []
+    }
+    pub fun resolveView(_ type: Type): AnyStruct? {
+        return nil
+    }
+}
+```
+We propose that all metadata be expressed in cadence types. The `pub fun availableViews(): [Type]`
```
turbolent, line 61:
```suggestion
We propose that all metadata be expressed in Cadence types. The `pub fun availableViews(): [Type]` 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Example two:
+This example shows how you might implement this specification with the ability to add new view types to your NFT 
+contract after it has already been published through the use of a `ViewResolver` pattern.
+[Gist available here](https://gist.github.com/briandilley/962e94682b7945d882fcd99702011ea4),
+[Playground available here](https://play.onflow.org/a71203b9-a03a-47f9-ab70-7285fcf4c56a?type=account&id=0)
+
+## Compatibility
+Provided support for default methods on interfaces is added to cadence there are no backwards compatibility issues. 
+  Short of that, NFT authors will need to add implementations for the new methods added to their contracts.
```
turbolent, line 304:
```suggestion
Provided support for default methods on interfaces is added to Cadence there are no backwards compatibility issues. 
Short of that, NFT authors will need to add implementations for the new methods added to their contracts.
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,320 @@
...
+## Engineering Impact
+
+Adhering to this standard will require contract authors to implement a few new methods as part of their NFT. 
+They will also likely need to import standard types into their contracts.
+
+There may also need to be a committee, process, and code repository put in place for managing the standard metadata 
+types.
+
+## Best Practices
+NFT Authors should do their best to implement as many views from the standard metadata set as appropriate for
```
turbolent, line 279:
```suggestion
NFT authors should do their best to implement as many views from the standard metadata set as appropriate for 
```

turbolent:
Nice work!

bjartek:
In this FLIP the method name `availableViews` is used but in my examples and reference impl i use the name `getViews`. What should we prefer?

bjartek:
I am not sure how to progress with merging changes here. But the latest work done here is here: 

Resource that contains shared types
- https://github.com/findonflow/find/blob/main/contracts/TypedMetadata.cdc

Proposed new NFT interface
- https://github.com/findonflow/find/blob/main/contracts/standard/NonFungibleToken.cdc

Implementation for Versus
- https://github.com/findonflow/find/blob/main/contracts/Art.cdc

psiemens (commented):
This is mostly a semantic review. Great proposal overall!

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,399 @@
...
+			self.source=source
+			self.name=name
+			self.thumbnail=thumbnail
+			self.description=description
+		}
+	}
+```
+
+The data here is taken from what blocto uses to display an NFT in their application.
+ - Would it be wise to add some sort of link here to be able to know how to navigate to the source page for this resource?
```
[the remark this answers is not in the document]
bjartek: I guess it is. We have not agreed on this in the work done thus far. 

This came up as I talked through this with my wife (she is into eth). TokenURI is a think there that is the uri to an NFT. Personally I think the Display struct should support a {String:String} map of links. One of them can be a tokenURI link.

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,399 @@
...
+
+// a file hosted by a web server with a hash for verifying it's contents
+pub struct URLWithHash {
+    pub let url: String
+    pub let hash: String?
+    pub let hashAlgo: String?
+}
+```
+
+### Royalty
```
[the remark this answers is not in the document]
bjartek: So you think we should just remove it from the flip and add it to the work we do in the NFT repo?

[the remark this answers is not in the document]
bjartek: Removed versus. 

The royalty interface is the one view that we actually have had a meeting on and agreed on so I feel that should be in there. We could just remove some of the more detailed Media examples if we feel it is still to large?

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,354 @@
...
+an argument.
+
+This proposal by no means precludes the use of member variable on the NFT resource itself. In fact,
+ideally all immutable data for an NFT is stored as immutable members of the NFT resource and views 
+are resolved and created by accessing those immutable members.
+
+### Supporting other resources besides NFTs
+In order to support any other resource and resource collections, like marketplaces and fungible tokens, the following new interface is proposed:
+```
+pub resource interface ViewResolverCollection {
```
briandilley, line 85: should this implement `ViewResolver` too?  It's not very useful as is, the IDs are basically irrelevant to anyone using this - they would have to iterate through every resolve in the collection to find one that supports the `Type` that they are trying to get a view for. If this instead implemented `ViewResolver` then all of that could happen internally and potentially have a more optimized implementation (more optimized than being forced to iterate until you find one that supports your desired view).

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,354 @@
...
+This will allow any collection to be exposed as view resolvable.
+
+For single resources that are stored as is like a fungible token or a profile contract you can just implement the `ViewResolver` interface directly.
+
+### Converting between different views
+
+Standards will evolve and new views will be added. In order to make it easier for maintainers we propose a mechanism to register a converter from one View to another:
+
+```
+pub struct interface ViewConverter {
```
briandilley, line 100: Not sure if this is very useful. You wouldn't be able to convert _anything_ to _anything_ else (ie: converting an `Animal` to a `IPFSImage`), and in the case that you're trying to convert `A` to `B` you could instead just look for a `ViewResolver` that supports `B` and ask for the `B` directly rather than finding a `ViewResolver` for `A`, then iterating through `ViewConverter`s until you found one that converts `A` to `B` and then converting it.

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,354 @@
...
+ 
+### Basic Views
+
+The most basic example is having `String` be a view. Think of this as `toString()` in Java or a `Stringer` implementation in Go. It is a way to view your NFT as a string. 
+
+The bigger brother of `String` would be `Display`. Here is how you could display an NFT in a wallet, marketplace or other user-facing application:
+```
+pub struct Display{
+    pub let name: String
+    pub let thumbnail: String
```
briandilley, line 119: should `thumbnail` be a `URLWithHash` instead? (ie: a URL to an image and a hash to verify it's contents).

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,399 @@
+# Non Fungible Token Metadata
+
```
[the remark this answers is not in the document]
bjartek: done

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,354 @@
...
+		pub fun calculateRoyalty(type:Type, amount:UFix64) : UFix64?
+
+		/// call this with a vault containing the amount given in calculate royalty and it will be distributed accordingly
+		pub fun distributeRoyalty(vault: @FungibleToken.Vault) 
+
+		/// generate a string that represents all the royalties this NFT has for display purposes
+		pub fun displayRoyalty() : String?  
+
+	}
+```
```
bluesign, line 289: I think this is little problematic to have `calculateRoyalty` and `distributeRoyalty` here, wouldn't it let to contract owner to update contract and then get all sale revenue to himself?  

or can change artist royalty address etc ? 

So far we addressed things to be usually read-only, but this is a big leap

flips/20210916-nft-metadata.md
bjartek: My thinking has been here that all NFTs in the same collection would share the same views. But i can see where that might not be the case. 

Maybe add a method that allow you to fetch all ids that have a given view?

flips/20210916-nft-metadata.md
bjartek: If you register a function that knows how to convert an Animal to an IPFSImage then you would be able to.

The use case here is if you have an NFT that knows how to resolve A and then a function from A -> B. So then you can also resolve that NFT as B via the ViewConverter.

flips/20210916-nft-metadata.md
bjartek: I have no problem with that, but I think the correct place for this discussion is in the NFT repo where we suggest the Display struct.

flips/20210916-nft-metadata.md
bluesign: As @briandilley said `ViewResolver` can be better, so it can return `[Display]` for example as supported type, and when queried, can return resolved for each item. ( though in that case gas limit can be a problem )

flips/20210916-nft-metadata.md
bjartek: This suggestion was made by Dieter. So I think adding this to the relevant place in the NFT repo for discussing views is the right thing to do.

flips/20210916-nft-metadata.md
briandilley: But why not just resolve the `B` view and cut out the conversion?

briandilley: WIth this in place, when I want to get a view of type `B` i have to first check the view resolver methods to see if there is a resolver for `B`, and if not - I then have to cycle through all of the `ViewConverter`s to find one that converts to `B` from another type that I also know. It just seems like a lot of steps vs. just being able to resolve the `B` view.

flips/20210916-nft-metadata.md
briandilley: Because `ViewResolver` is an interface, it could basically be implemented in any way the programmer wants - I'm not seeing the benefit of the `ViewResolverCollection` (at least as part of the standard anyway).

flips/20210916-nft-metadata.md
bluesign: I think here also can use `ViewResolver` maybe, instead of `to` and `convert`. 

If I have `ArtistConverter` instead of `convert` function, it can just have a `ViewResolver` that can provide destination formats.

flips/20210916-nft-metadata.md
bjartek: Without ViewResolverCollection we have no way of reasoning over collections of resources that are not NFTs where the resources share the same views. As an example NFTStorefront.

bluesign: I think gasLimit on large collections is a big concern too, you cannot query 1000 items for a Type.

flips/20210916-nft-metadata.md
bluesign: > WIth this in place, when I want to get a view of type B i have to first check the view resolver methods to see if there is a resolver for B, and if not - I then have to cycle through all of the ViewConverters to find one that converts to B from another type that I also know. It just seems like a lot of steps vs. just being able to resolve the B view.

if you want B, I think you will have a list of converters that converting to B, you will get a set of `from`s, then check if NFT resolves any of them I guess.

bluesign: I think this is more like: Imagine TopShot not supporting Blocto's `[BloctoDisplay]` view , so blocto can write a converter, use TopShot view then converter to get `[BloctoDisplay]`

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+}
+
+// a location on the planet
+pub struct GeoPoint {
+    pub let lat: UFix64
+    pub let lng: UFix64
+    pub let projection: String
+}
+
+// a file hosted by a web server with a hash for verifying it's contents
```
rudi-cilibrasi, line 245: this `it's` should be `its`

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+| **FLIP #**    | [636](https://github.com/onflow/flow/pull/636)                                                                                       |
+| **Author(s)** | Bjarte Karslen (bjartek@find.xyz), Deniz Mert Edincik (Deniz@edincik.com), Brian Dilley (briandilley@briandilley.com) |
+| **Sponsor**   | Peter Siemens (peter@dapperlabs.com)                                                                                                 |
+| **Updated**   | 2021-12-06                                                                                                                           |
+
+
+## Objective
+This proposal will make it possible to create generic solutions that will interoperate 
+through views and not hard-coded NFT types. Want to create a marketplace? If you know 
+how to display a certain Type that defines a sellable Item you can do so. And it will
```
louisguitton, line 14:
```suggestion
how to display a certain Type that defines a sellable Item, you can do so. And it will 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+## Objective
+This proposal will make it possible to create generic solutions that will interoperate 
+through views and not hard-coded NFT types. Want to create a marketplace? If you know 
+how to display a certain Type that defines a sellable Item you can do so. And it will 
+work for any contract and any resource, not just a single one. 
+
+## Motivation
+Currently there is no standard for defining NFT/resource metadata. At best, most implementations 
+employ a simple `{String: AnyStruct}` dictionary. Due to the various methods that NFT 
+authors use it’s hard / impossible to create applications that use NFTs whether it be
```
louisguitton, line 20:
```suggestion
authors use, it’s hard / impossible to create applications that use NFTs whether it be 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+work for any contract and any resource, not just a single one. 
+
+## Motivation
+Currently there is no standard for defining NFT/resource metadata. At best, most implementations 
+employ a simple `{String: AnyStruct}` dictionary. Due to the various methods that NFT 
+authors use it’s hard / impossible to create applications that use NFTs whether it be 
+for something as simple as viewing them or as complex as integrating them into a game 
+or application.
+
+If you want to create an Auction contract on Flow today you have to hardcode all the NFTs
```
louisguitton, line 24:
```suggestion
If you want to create an Auction contract on Flow today, you have to hardcode all the NFTs 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+authors use it’s hard / impossible to create applications that use NFTs whether it be 
+for something as simple as viewing them or as complex as integrating them into a game 
+or application.
+
+If you want to create an Auction contract on Flow today you have to hardcode all the NFTs 
+you want to support. This will often lead to duplicated contracts and code and just lots 
+of things to maintain. This standard will enable you to create a single Auction contract 
+and define the Type of views you need to expose in your NFT in order to support the contract. 
+So it will enable a design-by-contract pattern where the creator of a contract can define 
+the views it requires and an NFT can tell what views they export.
```
louisguitton, line 29:
```suggestion
the views it requires and a NFT can tell what views they export. 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+you want to support. This will often lead to duplicated contracts and code and just lots 
+of things to maintain. This standard will enable you to create a single Auction contract 
+and define the Type of views you need to expose in your NFT in order to support the contract. 
+So it will enable a design-by-contract pattern where the creator of a contract can define 
+the views it requires and an NFT can tell what views they export. 
+
+All NFTs and contracts using NFTs on Flow are affected by this problem. This is most 
+visible in the number of contracts that Viv3 needs in order to support all their different 
+sellable items. With all the oncoming marketplaces/auction houses and things like that 
+coming to Flow a good standard for NFTs will help everybody.
```
louisguitton, line 34:
```suggestion
coming to Flow, a good standard for NFTs will help everybody. 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+sellable items. With all the oncoming marketplaces/auction houses and things like that 
+coming to Flow a good standard for NFTs will help everybody. 
+
+## User Benefit
+We feel that a standard that authors can implement that enables easy consumption of metadata 
+will lead to more applications that integrate NFTs. This will make NFTs intrinsically more 
+useful and valuable to the end users holding the NFTs.
+
+A user will be much more likely to transfer their NFT to other marketplaces and sell their NFT 
+in different ways with this solution. Without it all marketplaces need to hardcode support
```
louisguitton, line 42:
```suggestion
in different ways with this solution. Without it, all marketplaces need to hardcode support 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+and define the Type of views you need to expose in your NFT in order to support the contract. 
+So it will enable a design-by-contract pattern where the creator of a contract can define 
+the views it requires and an NFT can tell what views they export. 
+
+All NFTs and contracts using NFTs on Flow are affected by this problem. This is most 
+visible in the number of contracts that Viv3 needs in order to support all their different 
+sellable items. With all the oncoming marketplaces/auction houses and things like that 
+coming to Flow a good standard for NFTs will help everybody. 
+
+## User Benefit
```
louisguitton, line 36: FLOW developers' time will be saved, they won't have to integrate N contracts into their application, meaning they can use their dev cycles for other things that will indirectly benefit the user.

Is it worth mentioning this indirect user benefit?

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+It will also open up the possibility to standardize on some shared structs like Royalty 
+and SaleItem/Auction.
+
+## Design Proposal
+
+The core of this proposal is to add the following interface and have the NFT resource implement it
+```
+pub resource interface ViewResolver {
+  pub fun getViews() : [Type]
+  pub fun resolveView(_ view:Type): AnyStruct
```
louisguitton, line 54:
```suggestion
  pub fun resolveView(_ type: Type): AnyStruct
```
below the naming is different

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+}
+```
+
+We will also add default implementations for these methods so that no compatibility will be broken 
+for existing NFT solutions. The new NFT interface could look something like this:
```
louisguitton, line 59:
```suggestion
for existing NFT solutions. The new NFT resource could look something like this: 
```
or do you mean `INFT`?

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+We propose that all metadata be expressed in Cadence types. The `pub fun getViews(): [Type]` 
+method allows for discovery of the various types of metadata supported by the NFT by returning a `Type` 
+for each supported type of metadata. An instance of any supported type could then be retrieved by 
+calling the `pub fun resolveView(_ type: Type): AnyStruct` method passing the desired type in as 
+an argument.
+
+This proposal by no means precludes the use of member variable on the NFT resource itself. In fact,
+ideally all immutable data for an NFT is stored as immutable members of the NFT resource and views
```
louisguitton, line 79:
```suggestion
ideally, all immutable data for an NFT is stored as immutable members of the NFT resource, and views 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+
+An on-chain image:
+data: data:image/png;base64,SOMEPNGDATAURI/wD/
+contentType: image/jpeg
+protocol: data
+
+The problem here is that the on-chain image can be quite large and you might not want to resolve those every time you fetch the media view. So what if we made a method on the struct that allowed you to retrieve the data/content instead?
+
+
+The same NFT author may also want expose other metadata about the artwork:
```
louisguitton, line 176:
```suggestion
The same NFT author may also want to expose other metadata about the artwork:
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+pub struct URLWithHash {
+    pub let url: String
+    pub let hash: String?
+    pub let hashAlgo: String?
+}
+```
+
+### Royalty 
+
+This particular problem has been discussed extensivly. The following suggestion was agreed upon
```
louisguitton, line 255:
```suggestion
This particular problem has been discussed extensively. The following suggestion was agreed upon
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+by @bjartek, @rheaplex and @dete. 
+
+```
+  /// a struct interface that describes how to calculate royalty 
+	pub struct interface Royalty {
+
+		/// if nil cannot pay this type
+		/// if not nill withdraw that from main vault and put it into distributeRoyalty
```
louisguitton, line 263:
```suggestion
		/// if nil cannot pay this type
		/// if not nil withdraw that from main vault and put it into distributeRoyalty 
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+
+```
+  /// a struct interface that describes how to calculate royalty 
+	pub struct interface Royalty {
+
+		/// if nil cannot pay this type
+		/// if not nill withdraw that from main vault and put it into distributeRoyalty 
+		pub fun calculateRoyalty(type:Type, amount:UFix64) : UFix64?
+
+		/// call this with a vault containing the amount given in calculate royalty and it will be distributed accordingly
```
louisguitton, line 266:
```suggestion
		/// call this with a vault containing the amount given in, calculate royalty and it will be distributed accordingly
```

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+
+		/// generate a string that represents all the royalties this NFT has for display purposes
+		pub fun displayRoyalty() : String?  
+
+	}
+```	
+
+## Drawbacks
+This standard does not include any required ways to store the metadata, and therefore some of the metadata 
+could be dynamic depending on how the NFTs are authored. This is a double edged sword; it allows for greater
```
louisguitton, line 277:
```suggestion
could be dynamic, depending on how the NFTs are authored. This is a double edged sword; it allows for greater 
```

louisguitton (approved):

flips/20210916-nft-metadata.md
bjartek: It could be, can you make a suggestion?

flips/20210916-nft-metadata.md
bjartek: I think this should be view actually, so if it is different below we should fix it there.

flips/20210916-nft-metadata.md
bjartek: Can you make a suggestion for this and then i will apply it @rudi-cilibrasi ?

flips/20210916-nft-metadata.md
```diff
@@ -0,0 +1,338 @@
...
+will lead to more applications that integrate NFTs. This will make NFTs intrinsically more 
+useful and valuable to the end users holding the NFTs.
+
+A user will be much more likely to transfer their NFT to other marketplaces and sell their NFT 
+in different ways with this solution. Without it, all marketplaces need to hardcode support 
+for all NFTs directly. 
+
+It will also open up the possibility to standardize on some shared structs like Royalty 
+and SaleItem/Auction.
+
```
louisguitton, line 47:
```suggestion

Indirectly, users will benefit from the fact that Flow developers can put their development cycles to better use than integrating each new NFT contract to their platform.
```

flips/20210916-nft-metadata.md
rudi-cilibrasi: it's fixed already ty

bjartek: yeah i fixed it :)

flips/20210916-nft-metadata.md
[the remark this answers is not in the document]
dete: I just want to make sure we can separate the discussion about how royalties should be handled from the discussion about the base View representation. If the Royalty example isn't contentious, we can certainly leave it here, but I don't want disagreement about how royalties should be handled to block adoption of the core proposal here. Specific views (like all of the examples here) can and should be discussed in new threads specific to those issues.

flips/20210916-nft-metadata.md
psiemens: Looks like this was resolved below, nice 👍

flips/20210916-nft-metadata.md
psiemens: Looks like this is resolved, so going to collapse this thread.

flips/20210916-nft-metadata.md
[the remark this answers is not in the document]
psiemens: @bjartek I'd like to propose that we remove the royalty view from this FLIP and put it in a new PR on the NFT repository (where this standard will live).

I know that you, @rheaplex and @dete talked about it, but I think we need to gather some more feedback before we solidify this part of the standard, specifically from the other marketplace teams (e.g. Rarible, BloctoBay, OpenSea) that would be building interfaces on top of the royalty view.

For example, here are a couple points I'd like to bring up in a separate thread:
- How does this view fit with the `NFTStorefront` contract?
- Similar to @bluesign, I'm worried about how opaque this interface is. I'm sure there was a tradeoff made at some point, but that tradeoff should be communicated alongside the proposed interface.

I'm committed to moving quickly on this (i.e. this week) but I think we're almost certainly going to delay the merging of this PR if we try to do it here.

psiemens (approved):
This looks good to go.

Thanks again @bjartek @briandilley and @bluesign!

dete (approved):
I don't see any remaining issues that should block this from proceeding. Great work, everyone!

dete (changes requested):
Oh shoot! I just noticed that the return type on `resolveView` doesn't have the `?` to make it optional. I think this is preferred!

bjartek:
> Oh shoot! I just noticed that the return type on `resolveView` doesn't have the `?` to make it optional. I think this is preferred!

I vaguely remember we talked about this and that this would be preferred to panicing the resolveView when you called it with a unknown view. I will fix it.

dete (approved):
👍

psiemens:
I'm now merging this PR and will update the status to "accepted". Thanks again, everybody!

The contents of this FLIP will be soon added to the [onflow/flow-nft](https://github.com/onflow/flow-nft/) repository alongside the base NFT interface.

PR #641, opened 2021-09-23
Update and add docs about secrets database

jordanschalm:
⚠️ Must be merged in tandem with a `boot-tools` release (https://github.com/onflow/flow-go/pull/1340) ⚠️ 

## Description

* Update existing `boot-utils` output examples to include generated encryption key
* Add guide for generating a database encryption key for existing operators
* Add information about the secrets database


docs/content/node-operation/node-bootstrap.mdx
```diff
@@ -38,6 +38,8 @@ curl -sL -O storage.googleapis.com/flow-genesis-bootstrap/boot-tools.tar
 tar -xvf boot-tools.tar
 ```
 
+$$$$$ TODO UPDATE BEFORE MERGING $$$$$$
```
jordanschalm, line 41: TODO

docs/content/node-operation/db-encryption-existing-operator.mdx
```diff
@@ -0,0 +1,55 @@
...
+Follow the instructions [here](/node-operation/node-bootstrap/#download-the-bootstrapping-kit)
+to download the latest version of the bootstrapping kit, then return to this page.
+
+## Generate Database Encryption Key
+
+You will need to generate an encryption key for the database using the `bootstrap` utility.
+
+<Callout type="warning">
+
+  Ensure you run the following commands on the machine you use to run your node software.
```
vishalchangrani, line 33: Will the script complain if the directory passed with the `-o` option not the boostrap directory. On a call with BisonTrail yesterday, they told use that their key generation process is kept separate from the node and for the machine account key generation they had to modify the script to not fail if the node id file is not found.

vishalchangrani (approved):
lgtm - but had a question

docs/content/node-operation/db-encryption-existing-operator.mdx
jordanschalm: No it won't work  -- this works the same way as the machine account generation.

jordanschalm: Do you think we should change the script? The idea behind that requirement is that the script also moves the file in the correct location. 

I just tested, and another workaround they could use is to create a `node-id` file in the environment they use for key generation, then run the script.
```
mkdir fakebootstrap
mkdir fakebootstrap/public-root-information
echo $NODEID > fakebootstrap/public-root-information/node-id
./boot-tools/bootstrap db-encryption-key -o fakebootstrap
```

kc1116 (approved):

huitseeker (commented):
Awesome possum. I would add neon signs to the "### Confidential Data & Files" paragraph, if I could, but eh.

PR #660, opened 2021-10-07
FLIP: Variable transaction fees

janezpodhostnik:
## Description

First draft

______

For contributor use:


janezpodhostnik:
Updated to simplify how the transaction fees are split up.

The new proposed way is to split them int a *known* part and *unknown* part (for the lack of better words). The only way to the *unknown* part is to actually run the transaction.

flips/20211007-transaction-fees.md
```diff
@@ -55,151 +57,152 @@ From the users perspective the transaction fees implementation should satisfy th
...
 2. Transaction failed before the transaction script execution started (some non-payer signature was incorrect, or the sequence number was incorrect)
 3. Transaction failed during transaction script execution, during fee deduction or during the storage used check. 
-4. Transaction failed because the computation limit was reached.
+4. Transaction failed because the execution effort limit was reached.
 
-As discussed below, in scenario 1. the transaction shouldn't have been included in a collection in the first place. However since it was included and there is no way to charge the payer, the collection node account is charged with the transaction fees. The transaction fees are computed with the dynamic computation effort set to 0.
+As discussed below, in scenario 1. the transaction shouldn't have passed the access node in the first place. However since it has and there is no way to charge the payer, the access node account is charged with the transaction fees. The transaction fees are computed with the execution effort set to 0.
```
j1010001, line 141: I'd probably merge the proposal on the charges with the cases above, otherwise I am jumping back and forth between the different cases and  how we propose to charge for them.

flips/20211007-transaction-fees.md
```diff
@@ -55,151 +57,152 @@ From the users perspective the transaction fees implementation should satisfy th
...
 
-It might still happen that this transaction will get included, either due to a bug or malicious behaviour from the collection node, or due to the collection node not having up to date information of the balance of the payer. In case does still get included in a collection, even though it does not meet the condition, this check should be repeated in the execution node, and in case the check fails on the execution node, the transaction fees should be deducted from the collection nodes account instead of the payers account.
+It might still happen that this transaction will get included in a collection, either due to a bug or malicious behaviour from the access node, or due to the access node not having up to date information of the balance of the payer. In case does still get included in a collection, even though it does not meet the condition, this check should be repeated in the execution node, and in case the check fails on the execution node, the transaction fees should be deducted from the access nodes account instead of the payers account.
```
j1010001, line 157: I think we should be explicitly calling out that with this we propose that access nodes will need to handle the scenario where an account attacks the access node by submitting transaction that will drain the access node funds - for example by having the ability to charge the account that submitted the transaction via alternative payment methods.

j1010001 (approved):

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+
+### Calibrating the parameters
+
+When turning on variable transaction fees the goal is to have a smooth transition to the new system. To help facilitate that the cost of transaction fees should be set so that ~95% of transactions should actually pay less transaction fees with variable transaction fees turned on.
+
+To achieve this the 6 different parts of variable transaction fees need to be calibrated accordingly.
+
+- <img src="https://render.githubusercontent.com/render/math?math=E_i"> Inclusion effort would be the byte size of a transaction plus a constant inclusion effort that would be picked by using linear regression on the impact of different size transactions (with the same execution effort) on the network, and extrapolating for a 0 size transaction.
+- <img src="https://render.githubusercontent.com/render/math?math=f_i">: Effort cost factor would be picked so that for 95% of transactions that have been seen so far the inclusion fees would cost less then `TC/2`.
```
AlexHentschel, line 166: regarding parameter calibration: 
* so we have effort, e.g. we could define that every 1kb transaction size equals 1 unit of inclusion effort
* and then we have a effort cost, i.e. how much does 1kb transaction size costs (?)

In economy, you often draw a price vs quantity diagram: 
Screen Shot 2021-11-25 at 8 59 49 PM

I have introduced several quantities here:
* `E_95` the 95% effort percentile of current transactions (i.e. their size in kb)
* `F_c` the current (fixed) transaction fee
* Furthermore, lets consider another transaction A with different effort (This could be a very small transaction with close to minimal effort `E_min`, as shown above) or a larger transactions (illustrated below).
* Assuming you desire a linear cost-effort relationship:
   * this is a linear function which is uniquely defined by a single point `X` and a slope `α`
   * we already have one point fixed: `X_95 = (E_95, F_c)`
   * for another transaction with effort `E_A` we still lack a meaningful price value `F_A`
* So how do we determine `F_A`? 
   * This is where regression comes into play, but I don't think the ordinate intersection point is particularly helpful here (you call this the "0-size transaction"). As an illustration, consider the two different price curves `p^(1)` vs `p^(2)` in the figure above. While they differ in their slope, i.e. in how much they charge per unit of effort (price per 1kb tx size), they are both reasonable pricing models.
   * I think for deciding which curve to pick, we should look at the throughput _limitations_ for some sample transactions. Roughly speaking, if we have two transactions A and B and we get twice as many transactions of type A through the system compared to transaction B, transaction B should cost twice as much. 
       * Formally, we require at least two benchmark transactions, A and B with their respective size `E_A` and `E_B`
       * We could load our benchmarking system with only these transactions up to the limit and measure the maximally achievable throughput. By comparing the throughput limitations, we can infer the slope. If we use more than 2 transactions, we get a sample of possible slopes. 

Screen Shot 2021-11-25 at 9 13 16 PM

AlexHentschel (commented):
Thanks @janezpodhostnik for the detailed write up. Great, that really enables a systematic discussion around which pricing behaviour we desire. 
While reading, I found that the mathematical model is a bit hard to follow at times. I think that is due to varying notation and terminology on the one hand and a somewhat unusual mathematical formulation on the other hand. While from a purely mathematical perspective, a variety of formulations are possible to describe the same behaviour, some formulations are more intuitive for as humans than others. 
I have made detailed comments on where I would suggest to revise the mathematical model to better align with common intuitions and terms. Another point which deserves more discussion (and potentially refinement) is how we chose a suitable parameter setting to induce the desired client behaviour. I generally agree that some form of regression will be useful; though I don't quite understand the detailed proposition around "extrapolating for a 0 size transaction". I think we should probably have a follow-up discussion (in person) around this.

flips/20211007-transaction-fees.md
```diff
@@ -55,151 +57,152 @@ From the users perspective the transaction fees implementation should satisfy th
...
 
-The effort cost factors can be used in three different ways:
+The effort cost factor can be used in three different ways:
 
-1. The main usage of the effort cost factors is adjusting them when the network is under load. If the execution or verification nodes are struggling to execute all of the transactions, the computation effort cost factor should be increased. If the network finds it difficult to create collections from all the transactions coming in, but the execution nodes are still fine, the computation effort cost factor should be increased. 
+1. The main usage of the effort cost factors is adjusting it when the network is under load. 
 
-2. The second reason to tweak the effort cost factors is that over time some parts of the code will get optimized. This could cause the same amount of effort to be easier to handle. An example would be that if execution of transactions became 10% faster because of optimizations, the computation effort cost factor could be reduced by 10%.
+2. The second reason to tweak the effort cost factors is that over time as hardware improves, the same amount of effort will be easier to handle by the nodes.
```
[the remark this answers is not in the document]
janezpodhostnik: Yes, that is the sentiment I was going for, but I did not know how to word it. Thank you.

flips/20211007-transaction-fees.md
```diff
@@ -55,151 +57,152 @@ From the users perspective the transaction fees implementation should satisfy th
...
 
 ### Inclusion effort
 
-For the first iteration the dynamic inclusion effort can be defined to be equal to the byte length of the (rlp encoded) transaction. The maximum dynamic inclusion effort would be equal to the maximum allowed transaction size, which is currently 1,500,000 bytes.
+Inclusion effort can represent many different parts of the system. Listed below are some potential factors that may play into the inclusion effort part of the fees listed by node type. Adding factors to this list (and actually implementing them) can happen gradually. One way to notice additional factors that might play into the inclusion effort is to compare transactions that have similar inclusion effort but cause a different load on the system.
 
-The constant inclusion effort could be chosen by observing the impact of transactions with a different byte size and a similar computation effort. This data could be used to predict how much impact a transaction that is 0 bytes long would make.
+1. Access node:
+    1. Sending a transaction causes load on the access node (might depend on the byte size of the transaction).
```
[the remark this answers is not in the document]
janezpodhostnik: "The access node needs to potentially do a check (at a constant effort)"  this statement is incomplete. My focus probably jumped here. The statement is supposed to be: "The access node needs to potentially do a check that the payer can pay (at a constant effort)". No this check does not exist, I will try to make that clearer. 

This list is presented as a list of "potential factors" (stated in the paragraph above the list).

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+
+While the effort cost factor could be adjusted manually from the start, that would mean responding to network load would be very slow. A potential solution to adjusting the effort cost factor automatically would be a service that monitored the networks performance and sent a transaction to update the fee factors when there was a change in the network load.
+
+### Inclusion effort
+
+Inclusion effort can represent many different parts of the system. Listed below are some potential factors that may play into the inclusion effort part of the fees listed by node type. Adding factors to this list (and actually implementing them) can happen gradually. One way to notice additional factors that might play into the inclusion effort is to compare transactions that have similar inclusion effort but cause a different load on the system.
+
+1. Access node:
+    1. Sending a transaction causes load on the access node (might depend on the byte size of the transaction).
+    2. The access node needs to potentially do a check (at a constant effort).
```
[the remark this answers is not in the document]
janezpodhostnik: "The access node needs to potentially do a check (at a constant effort)" this statement is incomplete. My focus probably jumped here. The statement is supposed to be: "The access node needs to potentially do a check that the payer can pay (at a constant effort)".

As for the access node checking for the syntactical correctness of the cadence code, I have some thoughts:
- I checked and currently the access node does check that the script is syntactically correct, so this should be added to the list of things the access node needs to do, but the cost of this might be difficult to determine, without actually checking for syntactical correctness.
- What motivates the access node to actually do this? Will it be slashed if it submits an invalid transaction?
- The execution node has to re-perform this check anyway otherwise it cannot execute it, so this cost probably needs to be counted twice.

I will add it to this list so we can look into how costly checking for syntactical correctness and if it needs to be included in the effort calculation sooner or later.

flips/20211007-transaction-fees.md
janezpodhostnik: I switched the order of the chapters, so at the point when you are reading about the categorisation of failing transactions, you should already have knowledge about some transactions not being included due to the payer being unable to pay.

flips/20211007-transaction-fees.md
janezpodhostnik: I added more text about this, but the issue is still open ended.

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+
+This separation into a part that can be know before the transaction is executed, and a part that cannot be know before the transaction is executed, allows for better oversight from the user into how much the transaction will actually cost.
+
+Each part of the fees can be further split into *effort*(<img src="https://render.githubusercontent.com/render/math?math=E">) and an *effort cost factor*(<img src="https://render.githubusercontent.com/render/math?math=f">) of FLOW per unit of effort. The *effort* required for a specific transaction should be independent of when the transaction is sent[^1] and should only depend on the properties of the transaction. On the other hand the *effort cost factor* would change over time depending on network activity.
```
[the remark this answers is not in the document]
laynelafrance: "effort" is defined as cost of looking at the tx and running it?

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+    - <img src="https://render.githubusercontent.com/render/math?math=E_c">: Execution effort; The time invariant effort required to execute this transaction (run the script, process the results, send results to verification node, verify the results, ...). The Execution effort cannot be know without executing the transactions script.
+
+
+### Fee effort cost factor
+
+The effort cost factor <img src="https://render.githubusercontent.com/render/math?math=f"> should be defined on a smart contract and adjustable via the service account admin resource and should be accessible for everyone to read. 
+
+The effort cost factor can be used in three different ways:
+
+1. The main usage of the effort cost factors is adjusting it when the network is under load.
```
laynelafrance, line 89: In consideration of surge pricing, I think we should separate traffic as a unique parameter to charge for

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+
+The effort cost factor <img src="https://render.githubusercontent.com/render/math?math=f"> should be defined on a smart contract and adjustable via the service account admin resource and should be accessible for everyone to read. 
+
+The effort cost factor can be used in three different ways:
+
+1. The main usage of the effort cost factors is adjusting it when the network is under load. 
+
+2. The second reason to tweak the effort cost factors is that over time as hardware improves, the same amount of effort will be easier to handle by the nodes.
+
+3. A potential third use of the effort cost factor would be to keep the cost of transaction stable according to the USD value of flow. This would require a periodic job to run, check the USD value of FLOW and update the cost factor to account for any changes. This would allow for the average transaction fees to stay relatively close to a set USD value.
```
laynelafrance, line 93: Can we update to “fiat” value? Im trying to get us away from overfitting to US market

flips/20211007-transaction-fees.md
```diff
@@ -55,151 +57,152 @@ From the users perspective the transaction fees implementation should satisfy th
...
 
-While the effort cost factors could be adjusted manually from the start, that would mean responding to network load would be very slow. A potential solution to adjusting the effort cost factors automatically would be a service that monitored the networks performance and sent a transaction to update the fee factors when there was a change in the network load.
+While the effort cost factor could be adjusted manually from the start, that would mean responding to network load would be very slow. A potential solution to adjusting the effort cost factor automatically would be a service that monitored the networks performance and sent a transaction to update the fee factors when there was a change in the network load.
```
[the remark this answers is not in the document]
laynelafrance: > A potential solution to adjusting the effort cost factor automatically would be a service that monitored the networks performance and sent a transaction to update the fee factors when there was a change in the network load

+1

flips/20211007-transaction-fees.md
laynelafrance: Flagging that we should discuss whether or not this is managed at AN or LN (collection node) level. I think currently we're trending toward AN but I do think we should complete the discussion on how failing txs without payer are managed between those 2 node types. I know @AlexHentschel has many thoughts ;)

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+## Design requirements
+
+We can separate the requirements into two groups depending on perspective.
+
+From the protocols perspective the following requirements can be set:
+
+- Transactions that require more resources from the network to be processed, should cost more.
+- If the protocol is experiencing a lot of traffic, all transactions should become more expensive, until the high traffic subsides.
+
+The ultimate goal from the protocols perspective would be that transactions cost exactly the amount of FLOW needed to pay for the cost of the time needed by the nodes to process the transaction. However this would include so many variables that it would not be feasible. Instead the goal is to make a reasonably good approximation, and improve that approximation over time.
```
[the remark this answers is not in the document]
janezpodhostnik: > Does bandwidth refer to networking bandwidth, compute bandwidth, both, or neither?

Both, but primarily networking bandwidth. Not sure its worth elaborating on this here, as this is mostly a list of potential factors...

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+[^2]: Except when an explicit decision is made to change them.
+
+Transaction fees are deducted from the transaction payer's default Flow wallet automatically. This happens as the final step of transaction execution. If the transaction fails the fees are still deducted and no other state change (except the fee deduction) is committed.
+
+The transaction fees are collected on the [FlowFees](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc) smart contract and are used as part of the staking rewards. If there were more fees collected during an epoch then there were staking rewards, the leftover FLOW is kept in a pool for the next payout. This part would not be changed with this proposal, but it is important to note, as this means collecting more fees leads to less inflation or even deflation.
```
joshuahannan, line 29:
```suggestion
The transaction fees are collected in a FLOW Vault stored in the [FlowFees](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowFees.cdc) smart contract and are used as part of the staking rewards. If there were more fees collected during an epoch than the specified staking reward amount for that epoch, the leftover FLOW is kept in a pool for the next payout. This part would not be changed with this proposal, but it is important to note, as this means collecting more fees leads to less inflation or even deflation.
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+There is an existing concept of a execution limit (a.k.a.: gas limit or computation limit), but it is currently not tied to transaction fees. The execution limit is used only to prevent transactions from running for a very long time (or indefinitely). The current measurement of execution is also very rudimentary and often doesn't reflect how much time will the transaction take to execute.
```
joshuahannan, line 31:
```suggestion
There is an existing concept of a execution limit (a.k.a.: gas limit or computation limit), but it is currently not tied to transaction fees. The execution limit is used only to prevent transactions from running for a very long time (or indefinitely). The current measurement of execution is also very rudimentary and often doesn't reflect how much time the transaction will take to execute.
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+The motivation behind the first separation is to separate the fees into a part that can be known before executing the transaction script, inclusion fees, and a part that can only be know by executing the transaction script, execution fees. 
+
+The inclusion part of fees accounts for the resources needed by a transaction because of the transaction's properties (that can be known without executing the transaction's script). Transaction properties are things like the byte size of the transaction and the number of signatures. Some examples of resources that will be needed are the bandwidth to transfer the transaction from node to node, the verification of signatures and the verification of the sequence number.
+
+
+The execution part of fees accounts for the approximate operational cost of executing the transaction script, passing the results to the verification node and verifying the results. Execution fees cannot be know without actually executing the transaction script. In order to avoid having unexpectedly high execution fees, each transaction should specify a stopping limit for the execution fees.
```
joshuahannan, line 62:
```suggestion
The execution part of fees accounts for the approximate operational cost of executing the transaction script, passing the results to the verification node and verifying the results. Execution fees cannot be known without actually executing the transaction script. In order to avoid having unexpectedly high execution fees, each transaction should specify a stopping limit for the execution fees.
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+The inclusion fees and the execution fees can be further split into effort and an the cost of that effort in units of FLOW. The effort required for a specific transaction should be independent of when the transaction is sent[^1] and should only depend on the properties of the transaction. The cost of the effort required to execute a transaction can be a factor with the unit of FLOW per effort multiplied by the effort, however in general terms the price of effort should be a function of effort itself. This would allow increasing the price of effort if a lot of effort was used in a transaction.
```
joshuahannan, line 64:
```suggestion
The inclusion fees and the execution fees can be further split into effort and the cost of that effort in units of FLOW. The effort required for a specific transaction should be independent of when the transaction is sent[^1] and should only depend on the properties of the transaction.
The cost of the effort required to execute a transaction can be a factor with the unit of FLOW per effort multiplied by the effort, however in general terms the price of effort should be a function of effort itself. This would allow increasing the price of effort if a lot of effort was used in a transaction. 
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+The way to specify the stopping limit for the execution fees is to define a maximum execution effort.
+
+[^1]: Assuming there weren't any changes to the state that would make the transactions script execute differently.
+
+Finally we need to multiply the transaction fees by a (unit-less) surge factor that would be more than 1 in the case the network is experiencing heavy load.
+
+### Terminology
+
+- <img src="https://render.githubusercontent.com/render/math?math=F"> : Transaction fees.
```
joshuahannan, line 74:
```suggestion
- <img src="https://render.githubusercontent.com/render/math?math=F"> : Total transaction fees.
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+The maximum fees a transaction can incur can be written with the equation below. 
+
+<img src="https://render.githubusercontent.com/render/math?math=F%5E%7B%5Ctext%7Bmax%7D%7D%20%3D%20s%20%5Cleft%5B%20p_I(I)%20%2B%20p_E(E%5E%7B%5Ctext%7Bmax%7D%7D)%20%5Cright%5D">
+
+### Surge factor
+
+The surge factor serves only one purpose, which is to give all transactions on the network a surcharge or potentially a discount when the network load is high or low respectively. It should be defined on a smart contract and adjustable via the service account admin resource and should be accessible for anyone to read. 
+
+The surge factor could be adjusted manually from the start, by manually moniyoring the network and sending a transaction to update the surge factor. This would mean responding to network load would be very slow.
```
joshuahannan, line 95:
```suggestion
The surge factor could be adjusted manually from the start, by manually monitoring the network and sending a transaction to update the surge factor. This would mean responding to network load would be very slow. 
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+<img src="https://render.githubusercontent.com/render/math?math=F%5E%7B%5Ctext%7Bmax%7D%7D%20%3D%20s%20%5Cleft%5B%20p_I(I)%20%2B%20p_E(E%5E%7B%5Ctext%7Bmax%7D%7D)%20%5Cright%5D">
+
+### Surge factor
+
+The surge factor serves only one purpose, which is to give all transactions on the network a surcharge or potentially a discount when the network load is high or low respectively. It should be defined on a smart contract and adjustable via the service account admin resource and should be accessible for anyone to read. 
+
+The surge factor could be adjusted manually from the start, by manually moniyoring the network and sending a transaction to update the surge factor. This would mean responding to network load would be very slow. 
+
+A fist automated solution to adjusting the surge factor would be a service that would monitor the network load and send a transaction to update the surge factor when there was a change in the network load.
```
joshuahannan, line 97:
```suggestion
A first automated solution to adjusting the surge factor would be a service that would monitor the network load and send a transaction to update the surge factor when there was a change in the network load.
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+
+When automation will be in place for adjusting the surge factor, the surge factor could be adjusted frequently (in the span it takes to run a few blocks and detect a surge), but it should not change to drastically otherwise users cannot respond to the change.
```
joshuahannan, line 99:
```suggestion
When automation will be in place for adjusting the surge factor, the surge factor could be adjusted frequently (in the span it takes to run a few blocks and detect a surge), but it should not change too drastically otherwise users cannot respond to the change.
```

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+The effort cost factor can be used in three different ways:
+
+1. The main usage of the effort cost factors is adjusting it when the network is under load. 
+
+2. The second reason to tweak the effort cost factors is that over time as hardware improves, the same amount of effort will be easier to handle by the nodes.
+
+3. A potential third use of the effort cost factor would be to keep the cost of transaction stable according to the USD value of flow. This would require a periodic job to run, check the USD value of FLOW and update the cost factor to account for any changes. This would allow for the average transaction fees to stay relatively close to a set USD value.
+
+While the effort cost factor could be adjusted manually from the start, that would mean responding to network load would be very slow. A potential solution to adjusting the effort cost factor automatically would be a service that monitored the networks performance and sent a transaction to update the fee factors when there was a change in the network load.
+
```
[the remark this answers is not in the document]
joshuahannan: I think this makes sense, but would this apply to inclusion effort and execution effort, or just execution?

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,327 @@
...
+### Execution effort
+
+The execution effort represents the effort needed to execute the transaction script (the CPU time) and the effort to handle the execution results (transferring the necessary data to the verification node for the transaction to be verified). Because execution of the transactions script depends on the current execution state, it is impossible to definitively predict the execution effort before the actual on-chain execution of the transaction.
+
+Currently the execution effort is calculated by adding 1 for every loop or function call. This could be sufficient for the first implementation. In the future the calculation of the execution effort would be changed so that function calls would have a different execution effort cost assigned to them according to the time complexity of the method.
+
+Thi model should be replaced so that the execution effort better correlates with the time needed to execute the transaction. A possible replacement is to make some functions (that take longer to run) use more effort than others.
```
joshuahannan, line 145:
```suggestion
This model should be replaced so that the execution effort better correlates with the time needed to execute the transaction. A possible replacement is to make some functions (that take longer to run) use more effort than others.
```

joshuahannan (commented):
Great write-up! I mostly had typo fixes, but left a few other comments. Looking good so far! 😃

flips/20211007-transaction-fees.md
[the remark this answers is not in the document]
AlexHentschel: That would be a viable proposal, if such service is universally trusted. But who would be running such service for a fully decentralized system? How to we guarantee correct functioning of such service in a byzantine environment?

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,264 @@
...
+
+There are a few reasons why a transaction fails, which can be split into four categories, according to who finally gets charged for the transaction fees and how much do they get charged.
+
+1. Transaction failed because the payers signature was incorrect, or the payer does not have enough funds to cover maximum transaction costs (transaction costs with the execution effort set to the execution effort limit).
+2. Transaction failed before the transaction script execution started (some non-payer signature was incorrect, or the sequence number was incorrect)
+3. Transaction failed during transaction script execution, during fee deduction or during the storage used check. 
+4. Transaction failed because the execution effort limit was reached.
+
+As discussed below, in scenario 1. the transaction shouldn't have passed the access node in the first place. However since it has and there is no way to charge the payer, the access node account is charged with the transaction fees. The transaction fees are computed with the execution effort set to 0.
+
```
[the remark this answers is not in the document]
AlexHentschel: ### Follow-up summarizing discussion with Dete

It is a very subtle combination of: 
1. A variant of the verifier's dilemma on the execution nodes, where they stop checking for transaction duplicates because the collectors never produce duplicates. In this mind set, every conflicting sequence number is the fault of the payer. Checking for duplicates is a resource-intensive task. There is no mechanism to preventing execution nodes becoming lazy and falling for the Verifier's Dilemma. 
2. However, as soon as Execution Nodes stop checking, a byzantine collector cluster could double-charge any client account by duplicating transactions (forcing sequence number collisions without the client being at fault). 

The threshold of this attack is relatively high, as it requires all execution nodes to be lazy (though not intentionally malicious) _plus_ a byzantine collector cluster. Side note: Reaching 1/3 weight is sufficient to pull off a _single_ replication attack of every submitted transaction. To fully bleed an arbitrary account, a malicious supermajority in a cluster would be required.

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+    In this case, the fees are charged normally with the execution effort set to the actual usage before the error was produced.
+
+4. Transaction failed because the execution effort limit was reached.
+
+    For these transactions the payer is charged with the transaction fees where the execution effort is set to the execution effort limit.
+
+### Parameter calibration
+
+When turning on variable transaction fees the goal is to have a smooth transition to the new system. To help facilitate this, the cost of transaction fees should be set so that ~95% of transactions should actually pay less transaction fees with variable transaction fees turned on. The two different parts of variable transaction fees need to be calibrated accordingly.
```
AlexHentschel, line 209:
```suggestion
When turning on variable transaction fees, a smooth transition to the new system is desired. To help facilitate this, the cost of transaction fees should be set so that ~95% of transactions should actually pay less transaction fees with variable transaction fees turned on. The two different parts of variable transaction fees need to be calibrated accordingly.
```

AlexHentschel (approved):
very nice. Thanks for the thorough discussion of the topic. Math representation is consistent and clear. 👍  Looks great. 

Most of my comments are wording suggestions or to break a long sentence into two.

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+
+The access node can check the balance of the payer prior to including their transaction and if the balance is below the minimum account balance, it would not include the transaction. However this prevents the payer from topping up his own balance, after an hones mistake. Some thought is needed here to address this edge case.
+
+The problem of providing economic incentive to include transactions of not part of this FLIP. This FLIP only provides a way to discourage access nodes to include transactions that cannot be paid for by the payer. 
+
+### Failing transactions
+
+There are a few reasons why a transaction fails, which can be split into four categories, according to who finally gets charged for the transaction fees and how much they get charged.
+
+1. Transaction failed because the payers signature was incorrect, or the payer does not have enough funds to cover maximum transaction costs (transaction costs with the execution effort set to the execution effort limit).
```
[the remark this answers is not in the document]
AlexHentschel: cc @laynelafrance somewhat related to [your comment](https://github.com/onflow/flow/pull/660#discussion_r760729515)

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+
+Assuming an honest and aware payer, the payer would realize their transaction failed when their balance first dropped below minimum account balance. They would then top off their account before continuing to send more transactions.
+
+With an unaware payer (for example due to bad automation) or a malicious payer, that wants to drain the access node account or wants to cause excess traffic without paying transaction fees, it is in the access nodes best interest to not include these transactions.
+
+The access node can check the balance of the payer prior to including their transaction and if the balance is below the minimum account balance, it would not include the transaction. However this prevents the payer from topping up his own balance, after an hones mistake. Some thought is needed here to address this edge case.
+
+The problem of providing economic incentive to include transactions of not part of this FLIP. This FLIP only provides a way to discourage access nodes to include transactions that cannot be paid for by the payer.
```
justinbarry, line 185:
```suggestion
The problem of providing economic incentive to include transactions is not part of this FLIP. This FLIP only provides a way to discourage access nodes to include transactions that cannot be paid for by the payer. 
```

flips/20211007-transaction-fees.md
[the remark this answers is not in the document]
laynelafrance: * payer has enough funds
-> payer doesn't have enough funds, I think you meant?

and +1

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+### Transaction Fee estimation
+
+In order to estimate the minimum transaction fees needed for a transaction, some data needs to be retrieved from the state (via the access node):
+
+- The effort cost parameters.
+- The surge factor.
+
+The calculation of inclusion effort can be done locally. 
+
+If we assume execution effort is 0 we get the minimum possible transaction fees. The maximum possible transaction fees can be computed with the execution effort set to the chosen execution effort limit.
```
laynelafrance, line 252: For reference, can we share what the current computation limit is?

laynelafrance (approved):
Awesome, awesome document @janezpodhostnik - thank you so much for proposing this :pray:

bluesign:
Nice proposal, I know for sure you calculated, but what is the current situation with transaction fees vs staking rewards for nodes? How much percentage of the rewards coming from transaction fees? 

I am curious how transaction fees will scale as network scales. Economic side of things are pretty interesting.

joshuahannan:
@bluesign There is [an event that indicates](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowIDTableStaking.cdc#L904) how many rewards come from fees every week and last week, that number was a bout 13.9 FLOW, so currently that puts it at about 0.001% coming from fees currently, but that number should go up as we enable this fee structure and the chain gets more and more usage.
reference:
https://flowscan.org/transaction/df669cd5b615708e54d5589761906ca5137d4860f1f78a11ea3bed48ff458e82?eventTypeFilter=A.8624b52f9ddcd04a.FlowIDTableStaking.EpochTotalRewardsPaid

redallica:
How is the priority given to 2 different users aiming to get their transaction done ? Is it on a FCFS basis ? or other rules ? Hope it is not like Ethereum where it is basically designed as an auction where the one who pays more gets his transaction first.

j1010001:
> How is the priority given to 2 different users aiming to get their transaction done ? Is it on a FCFS basis ? or other rules ? Hope it is not like Ethereum where it is basically designed as an auction where the one who pays more gets his transaction first.

Without going into details and edge cases on how network latency and other network attributes affect the transaction ordering, we can say that the transaction ordering is based on the time a transaction was received. Also, the fee is determined by the network based on the resources your transaction consumes, so a user can’t give their transactions priority by offering to pay more.

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+In this document I used the term **effort**, **effort cost**, and **effort limit** instead of **gas**, **gas cost**, **gas limit**. As I feel these better describe the sentiment of what is going on.
+
+## Current design
+
+Currently transaction fees are the same for all transactions and don't change over time [^2]. The transaction fee amount is defined in the `FlowServiceAccount` smart contract as the `transactionFee` field (this can be seen [here](https://github.com/onflow/flow-core-contracts/blob/master/contracts/FlowServiceAccount.cdc) or [here](https://flow-view-source.com/mainnet/account/0xe467b9dd11fa00df/contract/FlowServiceAccount)).
+
+[^2]: Except when an explicit decision is made to change them.
+
+Transaction fees are deducted from the transaction payer's default Flow wallet automatically. This happens as the final step of transaction execution. If the transaction fails the fees are still deducted and no other state change (except the fee deduction) is committed.
+
```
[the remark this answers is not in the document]
janezpodhostnik: We are currently not checking this. The reason was that we wanted to allow a user with 0 funds to (during a single transaction) acquire funds (e.g. withdraw staking rewards) then use those to pay for the transaction fees.

In light of recent discussions the logic might change. If  the users balance is below the minimum account balance (which should be significantly larger than the maximum theoretical transaction fees) do not execute the transaction at all.

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+<!-- TODO: graph for the paragraph above! -->
+
+In the first iteration of variable transaction fees the effort cost functions can simply be constant coefficients: <img src="https://render.githubusercontent.com/render/math?math=F_E%20%3D%20p_E%20*%20E"> and <img src="https://render.githubusercontent.com/render/math?math=F_I%20%3D%20p_I%20*%20I">. The coefficients are referred to as the inclusion effort cost parameter and the execution effort cost parameter, or together as effort cost parameters.
+
+The effort cost parameters should be defined on a smart contract and adjustable via the service account admin resource and should be accessible for everyone to read.
```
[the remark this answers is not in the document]
janezpodhostnik: Yes the service account is the future governance account. 

It is also possible that the control of these specific parameters might be split out to be controlled by a different account.

Thats up to future decisions.

Right now the service account is the one in control of transaction fee rates and storage fee rates (and other things, like epoch parameters) and is a multisig account, so I think this currently fits there.

flips/20211007-transaction-fees.md
```diff
@@ -0,0 +1,329 @@
...
+
+<img src="https://render.githubusercontent.com/render/math?math=F%5E%7B%5Ctext%7Bmax%7D%7D%20%3D%20s%20%5Cleft%5B%20p_I(I)%20%2B%20p_E(E%5E%7B%5Ctext%7Bmax%7D%7D)%20%5Cright%5D">
+
+### Surge factor
+
+The surge factor serves only one purpose, which is to give all transactions on the network a surcharge or potentially a discount when the network load is high or low respectively. It should be defined on a smart contract and adjustable via the service account admin resource and should be accessible for anyone to read. 
+
+The surge factor could be adjusted manually from the start, by manually monitoring the network and sending a transaction to update the surge factor. This would mean responding to network load would be very slow. 
+
+A first automated solution to adjusting the surge factor would be a service that would monitor the network load and send a transaction to update the surge factor when there was a change in the network load.
```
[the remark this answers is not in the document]
janezpodhostnik: The suggestion to use a service to keep track of load and update the surge factor is a bad one and I will update this.

The reason it's a bad one is that that service would be external to the network, so it could not be trusted.

Your suggestion to keep track of block fullness is a lot better. I think I will rewrite this so that surge control logic is executed during the system transaction/chunk (at the end of every block).

PR #677, opened 2021-10-26
State sync engine design doc

smnzhu:
[Design doc for state sync engine](https://github.com/onflow/flow/blob/smnzhu/sync-engine-design/roadmap/state_sync_protocol/state-sync-protocol.md)

TODO:
- [x] Add diagrams


roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
+# Execution State Synchronization Protocol
+
+## Context
+
+Today, Access nodes act as the gateway to the Flow network, providing a query endpoint for users (REST, gRPC) and performing important admission control duties such as:
+* Rate limiting of requests (script executions, Collection requests, account queries, etc.)
+* Basic validation, preventing bad transactions from entering the network
+
+Execution nodes and Collection nodes currently still expose public APIs, vestiges of an earlier version of the protocol in which Access nodes had not yet been introduced. However, allowing users to bypass the Access layer and make direct requests to these nodes not only dilutes the valuable resources which should be reserved for performing tasks crucial to the protocol (especially in the case of Execution nodes), but it also leaves them at risk of denial-of-service attacks. Going forward, in order to protect the other staked nodes of the network, the Access layer will become the *only* way for users to interact with the network.
```
Kay-Zee, line 9: I feel like theres some inaccuracies here.

> vestiges of an earlier version of the protocol in which Access nodes had not yet been introduced

I'm not sure if this is the right way to think about this. Even after adding the access layer, we were still adamant about being able to interact with the chain through Collection and Execution nodes, as a feature. This section in the doc seems implies that it was an unwanted side effect.

> the Access layer will become the *only* way for users to interact with the network.

Also, this is quite contentious at the moment, so I'm not so sure it makes sense to have it here before a decision is made.

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+* [`GetTransaction`](/docs/content/access-api.md/#gettransaction) / [`GetTransactionResult`](/docs/content/access-api.md/#gettransactionresult) forwards the request to a Collection / Execution node and returns the response.
+* [`ExecuteScriptAtLatestBlock`](/docs/content/access-api.md/#executescriptatlatestblock) / [`ExecuteScriptAtBlockID`](/docs/content/access-api.md/#executescriptatblockid) / [`ExecuteScriptAtBlockHeight`](/docs/content/access-api.md/#executescriptatblockheight) forwards the request to an Execution node and returns the response.
+* [`GetAccountAtLatestBlock`](/docs/content/access-api.md/#getaccountatlatestblock) / [`GetAccountAtBlockHeight`](/docs/content/access-api.md/#getaccountatblockheight) forwards the request to an Execution node and returns the response.
+
+Ideally, Access nodes would be able to serve such requests from their own local state to avoid introducing additional traffic and putting excessive load on nodes within the network. To do this, they need the ability to maintain an up-to-date copy of the execution state and all transaction data.
```
Kay-Zee, line 18: :+1:

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,195 @@
...
+    repeated Event events = 3;
+    repeated TrieUpdate trieUpdates = 4;
+}
+```
+
+They will then serialize each of these fields<sup>[1](#footnotes)</sup> and compute a root CID from the serialized data as follows:
+
+1. Split the data into blobs of up to 1MB (the message size limit of the [underlying networking stack](https://docs.ipfs.io/concepts/libp2p/)).
+2. Compute the CID of each blob and store it in the blobstore.
+3. If the number of CIDs computed in the previous step is 1, then return this CID. Otherwise, concatenate the CIDs and repeat from step 1<sup>[2](#footnotes)</sup>.
```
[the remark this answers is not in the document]
Kay-Zee: Or at least add:

Otherwise, concatenate the CIDs and repeat from step 1 with the new concatinated data as a new blob

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+
+> There is a Verifier's Dilemma here which needs to be addressed. We will not go into the details of how to solve this here.
+
+#### Verification Nodes as Verifiers
+
+Since Verification nodes are each responsible for only a subset of the Chunks within a block, Execution nodes will need to compute Execution Data and CIDs for each individual Chunk. Once Verification nodes have the [Chunk Data Pack](https://github.com/onflow/flow-go/blob/e73e19a87860f470053b6a8b624d3dcd43ad23bf/model/flow/chunk.go#L40), they can reproduce the CIDs and check that they are correct.
+
+## Requirements and Rewards
+
+The Access node goes from being an optional node serving as a proxy, to being a relayer of state information. It should correspondingly offer SLAs on state re-transmissions and get rewarded for it.
```
Kay-Zee, line 116: I'd say the access node always had some aspects of "relaying" info, just use to be more protocol state related (and some misc data like events and result strings)

Also, as mentioned earlier, I believe discussions of the "optionality" of access nodes is still under discussion, so not sure if we want to include the idea here.

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+
+The root CID of the snapshot must be shared with new nodes somehow so that they can use it to download the snapshot via Bitswap. There are a few ways this can be done:
+
+* The CID can be shared on a broadcast channel using a *Byzantine consistent broadcast* algorithm<sup>[5](#footnotes)</sup>.
+* A new Access node can randomly select a couple of other Access nodes and simply ask them for the CID. If the responses it receives are concordant, it can go ahead and download the data. Once the download is complete, it can generate the state trie for each block and compare its hash to the one found on-chain in the corresponding Execution Result. If it finds an inconsistency, it has all the evidence it needs to submit a slashing challenge against the nodes which provided it with the bad CID.
+
+### Sharding
+
+The Execution Data can be sharded so that each Access node only needs to store and respond to requests for a fraction of the entire dataset.
```
Kay-Zee, line 180: somewhat similar to collection nodes and their clustering.

but yes, as mentioned later, this increased the severity of possible data availability issues, and should likely only be pursued after a critical mass of access nodes are part of the network

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
+# Execution State Synchronization Protocol
+
+## Context
+
+Today, Access nodes act as the gateway to the Flow network, providing a query endpoint for users (REST, gRPC) and performing important admission control duties such as:
+* Rate limiting of requests (script executions, Collection requests, account queries, etc.)
+* Basic validation, preventing bad transactions from entering the network
```
AlexHentschel, line 7: Technically, ANs can't _prevent_ bad transactions from entering the network. But they can provide a first level of filtering (imperfect but providing a significant hurdle to spamming the network).

roadmap/state-sync-protocol.md
AlexHentschel: > Execution nodes and Collection nodes currently still expose public APIs, vestiges of an earlier version of the protocol in which Access nodes had not yet been introduced.

Access nodes were already part of the architecture _before_ we started with the Go implementation. The reasons why we started with APIs on the ENs and Collectors was simply a short-cut. 

```suggestion
Execution nodes and Collection nodes currently still expose public APIs, vestiges of earlier short cuts. However, allowing users to bypass the Access layer and make direct requests to these nodes not only dilutes the valuable resources which should be reserved for performing tasks crucial to the protocol (especially in the case of Execution nodes), but it also leaves them at risk of denial-of-service attacks. Going forward, in order to protect the other staked nodes of the network, the Access layer will become the *only* way for users to interact with the network.
```

AlexHentschel: > the Access layer will become the only way for users to interact with the network.

agree with Kan. I think the sync engine will be needed either way, no matter whether the EN or the AN exposes the API. I would argue as follows:

> It is important for network-external entities as well as network-internal nodes to be able to acquire a full copy of the execution state. This is important to remove load from the Execution node. Furthermore, due to their (intentionally) low number, entities other than the Execution Nodes should be able to hold (and potentially archive) the full state to further decentralization.

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+* [`GetAccountAtLatestBlock`](/docs/content/access-api.md/#getaccountatlatestblock) / [`GetAccountAtBlockHeight`](/docs/content/access-api.md/#getaccountatblockheight) forwards the request to an Execution node and returns the response.
+
+Ideally, Access nodes would be able to serve such requests from their own local state to avoid introducing additional traffic and putting excessive load on nodes within the network. To do this, they need the ability to maintain an up-to-date copy of the execution state and all transaction data.
+
+The [Flow DPS](https://github.com/optakt/flow-dps) has been an important first step in this direction, providing a scalable and efficient way to access the history of the Flow execution state. It reads the protocol state by following block proposals via the [`ConsensusFollower`](https://github.com/onflow/flow-go/blob/master/follower/consensus_follower.go) library, and reads the execution-related data from records written to a Google Cloud Storage bucket by an Execution node. Going forward, we will want to incorporate this functionality into the Access node itself.
```
AlexHentschel, line 20: > The [Flow DPS](https://github.com/optakt/flow-dps) ...

Would suggest to write out the full name at the first occurrence: 


> The [Flow Data Provisioning Service [DPS]](https://github.com/optakt/flow-dps) ...

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+
+Bitswap functions with a few fundamental concepts:
+* A *block* represents a raw chunk of data.
+* Each block is uniquely identified by a [*content identifier (CID)*](https://docs.ipfs.io/concepts/content-addressing/), which is based on a cryptographic hash of the block's contents.
+* A *blockstore* is a data store for storing and retrieving blocks.
+* A *session* orchestrates the flow of messages with other peers for discovering and retrieving the individual blocks which comprise a single high-level object.
+
+The full details of the Bitswap protocol can be found [here](https://research.protocol.ai/publications/accelerating-content-routing-with-bitswap-a-multi-path-file-transfer-protocol-in-ipfs-and-filecoin/delarocha2021.pdf).
+
+> To avoid conflicting terminology, we will henceforth refer to blocks in Bitswap as *blobs*.
```
AlexHentschel, line 42: ❤️ 

formatting question: why is this a quote?

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+
+They will then serialize each of these fields separately<sup>[1](#footnotes)</sup> and compute a root CID from the serialized data for each field as follows:
+
+1. Split the data into blobs of up to 1MB (the message size limit of the [underlying networking stack](https://docs.ipfs.io/concepts/libp2p/)).
+2. Compute the CID of each blob and store it in the blobstore.
+3. If the number of CIDs computed in the previous step is 1, then return this CID. Otherwise, concatenate the CIDs and repeat from step 1<sup>[2](#footnotes)</sup>.
+
+**[INSERT DIAGRAM HERE]**
+
+At the end of this process, we are left with four root CIDs (one for each field in the Execution Data). These will be included in an [Execution Result](https://github.com/onflow/flow-go/blob/master/model/flow/execution_result.go) from which an [Execution Receipt](https://github.com/onflow/flow-go/blob/master/model/flow/execution_receipt.go) is generated and later broadcast to the network.
```
AlexHentschel, line 65: `Execution Results` _must_ be deterministic. If two ENs have the same result, they must generate the same `Execution Result`. This implies we need protocol definitions how elements `repeated TrieUpdate trieUpdates` are ordered.

roadmap/state-sync-protocol.md
smnzhu: > Also, this is quite contentious at the moment, so I'm not so sure it makes sense to have it here before a decision is made.

yeah, this was written before we had our previous discussion on Access nodes, and I haven't had time to update it since then

roadmap/state-sync-protocol.md
smnzhu: I think that *if* we end up deciding on having AN's be staked, and therefore the state sync protocol becomes part of the actual core protocol, then it's also possible to have Access Nodes do the verification, as described below in the Verification section.

On the other hand, if not, then yes verifier nodes will need to do the checking, and I have written a little bit on that in the Verification section as well. 

The high level idea is that instead of generating a single `ExecutionData` per block and creating a root CID(s) for that single `ExecutionData`, the Execution nodes would instead create a separate `ExecutionData` for each chunk. Therefore, instead of just one set of root CID's in the ExecutionResult, there would be `numberOfChunks` sets of root CID's. 

Then, when a verification node is verifying each chunk, they would recompute the `ExecutionData` for that chunk, and compare the root CID's they get vs the ExecutionResult's root CID's.

@AlexHentschel does this make sense? I will add this to the doc, if there are any suggestions on how to phrase this to make it more clear please let me know.

smnzhu: > This implies we need protocol definitions how elements repeated TrieUpdate trieUpdates are ordered.

Yes, thanks for that flag. My assumption was that there is already a defacto ordering (namely, the order that the updates were applied when executing the block), but maybe this is incorrect. Can you clarify here?

Once this is clear, I will add more detail to the doc.

roadmap/state-sync-protocol.md
```diff
@@ -0,0 +1,198 @@
...
+
+### Data Transfer
+
+The core data exchange protocol we build on top of is called [Bitswap](https://docs.ipfs.io/concepts/bitswap/). At a high level, Bitswap's main responsibilities are to acquire chunks of data requested by the client from the network, and send chunks of data in its possession to other peers who want them.
+
+Bitswap functions with a few fundamental concepts:
+* A *block* represents a raw chunk of data.
+* Each block is uniquely identified by a [*content identifier (CID)*](https://docs.ipfs.io/concepts/content-addressing/), which is based on a cryptographic hash of the block's contents.
+* A *blockstore* is a data store for storing and retrieving blocks.
+* A *session* orchestrates the flow of messages with other peers for discovering and retrieving the individual blocks which comprise a single high-level object.
```
zhangchiqing, line 38: There is existing meaning for "block" already, better to choose a different name to avoid confusing. For instance, "blob"
```suggestion
* A *blob* represents a raw chunk of data.
* Each blob is uniquely identified by a [*content identifier (CID)*](https://docs.ipfs.io/concepts/content-addressing/), which is based on a cryptographic hash of the blob contents.
* A *blobstore* is a data store for storing and retrieving blocks.
* A *session* orchestrates the flow of messages with other peers for discovering and retrieving the individual blobs which comprise a single high-level object.
```

roadmap/state-sync-protocol.md
smnzhu: yes, I should clarify and put "being a relayer of execution state information" here

smnzhu:
@AlexHentschel @Kay-Zee Thanks for the comments, this doc was written before we had the first discussion on whether access nodes should be staked or not.

In fact, writing this doc was one the reasons I began to have doubts about the Access node as a staked node, because I was having a hard time convincing myself on some of the points in this doc. I'll refactor the it to remove some of the discussion around staking-related topics.

@AlexHentschel In the meantime, it would be great if you could reply to [these](https://github.com/onflow/flow/pull/677#discussion_r748638090) followups.

roadmap/state-sync-protocol.md
smnzhu: Here I am using this formatting as more of a note: https://stackoverflow.com/a/43120795

Probably, it would make more sense if I added the "*_NOTE:_**" in the beginning though.

roadmap/state-sync-protocol.md
zhangchiqing, line 65: We serialize each field and make them as a CID, so in the end there will be 4 CIDs, since there are 4 fields. But if a field is bigger than 1MB, which exceeds the networking limit, we will further split it into smaller CIDs. So in the end, we could have more than 4 CIDs. right?

In this case, when I receive 5 CIDs in total, how do I know which 2 CIDs is more the field that is over 1MB and need to be combined?


```suggestion
At the end of this process, we are left with four root CIDs (one for each field in the Execution Data, or more if one field is bigger than 1MB). These will be included in an [Execution Result](https://github.com/onflow/flow-go/blob/master/model/flow/execution_result.go) from which an [Execution Receipt](https://github.com/onflow/flow-go/blob/master/model/flow/execution_receipt.go) is generated and later broadcast to the network.
```

zhangchiqing: Also, if a field is bigger than 1MB, how does the execution node let the access node know that there are 5 CIDs rather than 4 in total to be downloaded?

bluesign:
I reread the spec but I have some questions about economics ( keeping future decentralization in mind ) . 

Considering AN will be staked and will be rewarded for it and it will have 2 sides ( user facing and other AN facing ) 

**On the other AN facing side:** What will guarantee that AN will not be leecher and will answer other AN requests for bitswap? You have protection for bad response, but there is no challenge for no response. ( which I think also very tricky )

I think there has to be some kind of internal payment structure instead of reward structure. You can pay for to get some CID with some flow if response is bad, you can challenge and slash AN and get your Flow back (hopefully with some bonus)  from what is slashed. 

With this model, you can have reward and punishment, but also you can have some kind of healthy market for state data ( with ANs have ability to set their own prices). Technically you can make execution node to have higher price for CID request than other ANs.  Also ANs will compete to lowest price between each other.  

**On the user facing side:** What will guarantee AN to service user requests? ( This is not too critical as we already have the same risk now ) 

I think for user facing side, they should be compensated from transactions. Then they can even share this revenue with dapps. This will create a healthy market for ANs with their value added features ( like alchemy )

roadmap/state-sync-protocol.md
smnzhu: @zhangchiqing 

No, we will only have a single CID, because the CID's are arranged in a tree. 

I added a diagram in the latest version, does that help clarify things?

I would also love input on how to improve the diagram :)

AlexHentschel: > The high level idea is that instead of generating a single `ExecutionData` per block and creating a root CID(s) for that single `ExecutionData`, the Execution nodes would instead create a separate ExecutionData for each chunk.

👍  makes sense

>  defacto ordering (namely, the order that the updates occurred when executing the block)

The convention is that read and writes are deterministically ordered by their respective register ID. The ordering during execution is _part_ of the entropy that goes into creating the SPoCK proof. Therefore, we don't want to expose it as part of the result.

AlexHentschel:
@smnzhu this Flip has brought up some very interesting ideas:
* Per block, we have conceptually two consumers of the state updates:
  * The Verification Nodes, which receive their respective data through `ChunkDataPack`s
  * Full "observer nodes" (using this term in lack of a better one) that wish to obtain a copy of the full state. These nodes receive their updates over BitSwap
* There is a strong dependency between the information packages for the Verifiers and the Observers:
  * If we consider the block's computation one chunk at a time, the verifiers essentially need the input data (state and transactions) and reproduce the output data (state change, events, etc).
  * While there is some data (e.g. touched registers during a chunk computation incl. their initial values) that only the Verifiers need, other data (like collection full text) is also required by the Observers. 
  * As the verifiers also need to check the correctness of the commitment that the Observers use to verify their data, the Verifiers essentially need to re-create the same data structures that are consumed by the observers. In other words, when approving a chunk computation, the verifier also possesses the data that the Observers are interested in. 
* Furthermore, there also a strong overlap in requirements for the data transmission  EN -> Observer on the one hand and EN -> VN on the other hand:
  * For the observer, we use BitSwap specifically because it breaks the data into small blobs. Thereby, we keep the amount of work minimal that an Observer has to invest, before it can detect a potential attack from an EN feeding it garbage data. 
  * For Verifiers, we have the same requirement. But our implementation still sends large `ChunkDataPack`s (I think with a message size limit of 2GB!) which is not BFT compatible. 

From a conceptual perspective, it seems natural to unify the data exchange between ENs on the one hand and Verifiers plus Observers on the other hand: 
* we would remove the attack vector of huge message sizes against Verification Nodes 
* After verification, Verifiers will also posses much of the data for Observers
   * By sharing the same data structure, Verifiers could answer data requests. This significantly reduces the "data centralization" in flow towards a more decentralized ecosystem. 
   * In oder to fully avoid the Wintermute Attack, we need to slash Execution results that do not receive sufficient approvals. As a result, Execution Nodes have a strong interest in providing their data to the assigned Verifiers. In the mature protocol, there will be 50 - 80 Verifiers assigned to each chunk. If the verifiers could easily provide missing data to the Observers by sharing the same data format, we would (at least) 10x the number of staked parties that could provide the data of interest. 

     Note: there is still no intrinsic incentive in the protocol to motivate either the EN or VN to share their data. But increasing the number of parties by 10x at least makes a difference in case most participants are generally benevolent. 

In summary, my suggestion would be to evaluate feasibility:
1. [part of this FLIP] design level: lets work out the `ExecutionReceipts` data structure in more detail that would allow the VN and Observer to both retrieve their respective data of interest
2. [beyond this FLIP] If we can work out the details, I would like to propose this in our TPM. Chances are that this would actually sufficiently address our _practical_ concerns around data access. (fundamental challenges remain: lack of intrinsic incentive to share data and punishment for not doing so). 
3. [beyond this FLIP] Evaluate how to engineer this. Chances are that we:
   *  use the unified scheme first only for the state sync engine and still keep the `ChunkDataPack`s for some time until we updated the existing code.
   * If we are going down this route, we also need to separate the Bitswap network into a permissioned blockchain-internal network and an open external network.


A brief illustration below:

Screen Shot 2021-11-22 at 2 56 39 PM

smnzhu:
@Kay-Zee @AlexHentschel addressed most comments, please take one more pass :)

roadmap/state_sync_protocol/state-sync-protocol.md
```diff
@@ -0,0 +1,204 @@
+# Execution State Synchronization Protocol
+
+## Context
+
+Today, Access nodes act as the gateway to the Flow network, providing a query endpoint for users (REST, gRPC) and performing important admission control duties such as:
+* Rate limiting of requests (script executions, Collection requests, account queries, etc.)
+* Basic validation, proving a first level of filtering to prevent bad transactions from entering the network
+
+Execution nodes and Collection nodes currently still expose public APIs, which are now mostly obsolete. Allowing users to bypass the Access layer and make direct requests to these nodes not only dilutes the valuable resources which should be reserved for performing tasks crucial to the protocol (especially in the case of Execution nodes), but it also leaves them at risk of denial-of-service attacks. Going forward, in order to protect the other staked nodes of the network, the Access layer will become the *only* way for users to interact with the network.
```
smnzhu, line 9: > Going forward, in order to protect the other staked nodes of the network, the Access layer will become the *only* way for users to interact with the network.

After our latest meeting, are we still opposed to keeping this line?

smnzhu: > Execution nodes and Collection nodes currently still expose public APIs, which are now mostly obsolete.

What I mean here is that the *public* APIs are obsolete, meaning there is no need for these APIs to be public since external parties / clients can interact via the Access Node. Suggestions on how to phrase this to make it more clear are welcome.

tav:
Great spec!

My only concern is with the pruning. Would it be possible to exclude `Collections` from pruning? i.e. only prune events, tx results, and trie updates?

For operators who need to re-execute every transaction for auditing purposes, pruning suggests that they'll not be able to get the raw transactions after a certain number of blocks. If all collections were permanently archived however, it'd make life a lot easier for those operators.

roadmap/state_sync_protocol/state-sync-protocol.md
AlexHentschel: > Going forward, in order to protect the other staked nodes of the network, the Access layer will become the only way for users to interact with the network.

I think that this accurately represents the vision 👍

AlexHentschel:
@tav pruning should apply to collection and execution nodes, whose main role is to keep _extending_ the chain. Of course we desire that there is a way to retrieve archival data (including old transactions in full text). But what we are saying here is that the protocol will not intrinsically guarantee that arbitrarily old data is still available (which would be very hard to guarantee anyway). The access node software should allow the operator to retain old data (including collections) and make it available for querying. But we are not forcing the operator to do so.

roadmap/state_sync_protocol/state-sync-protocol.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+**Execution State Index (~3 week spork)**
+
+```console
+simon_zhu_dapperlabs_com@observer-001 /var/flow/data/mainnet-13 $ sudo du -sh index
+103G    index
+```
+
+Based on the numbers above, we should require Access nodes to have >43GB of RAM (27GB + current 16GB requirement) and minimum 1.5-2TB of storage space for a 3 month spork. These requirements can likely be decreased if sharding is implemented (see [Future Optimizations](#future-optimizations)).
```
AlexHentschel, line 168: I'd suggest to re-frame this and instead of sharding, talk about horizontal scaling:

```suggestion
Based on the numbers above, we should require Access nodes to have >43GB of RAM (27GB + current 16GB requirement) and minimum 1.5-2TB of storage space for a 3 month spork. However, these requirements don't have to be met by a single machine, because the historic state does not change. The historic state can be easily distributed over many physical machines allowing for arbitrary horizontal scaling (see [Future Optimizations](#future-optimizations)).
```

roadmap/state_sync_protocol/state-sync-protocol.md
```diff
@@ -0,0 +1,204 @@
...
+
+## Future Optimizations
+
+### Checkpointing
+
+Having a checkpointing mechanism will allow new nodes to catch up more quickly, while also reducing the amount of data Access nodes need to store on disk.
+
+Access nodes can periodically (e.g. at every 10000th block) generate a snapshot of the full state index and store it in their blobstore. They can then prune all of the Execution Data for blocks older than the snapshot, and require nodes that are behind to directly sync the snapshot first before catching up via per-block Execution Data updates.
```
AlexHentschel, line 176: Node operators might want to keep the historic state; pruning might not be desirable 

```suggestion
Access nodes can periodically (e.g. at every 10000th block) generate a snapshot of the full state index and store it in their blobstore. Thereby, new access nodes (and other external clients) could directly sync starting from the most recent snapshot and catch up via per-block Execution Data updates (before they sync the full state history if they desire to do so).
```

roadmap/state_sync_protocol/state-sync-protocol.md
```diff
@@ -0,0 +1,204 @@
...
+
+The root CID of the snapshot must be shared with new nodes somehow so that they can use it to download the snapshot via Bitswap. There are a few ways this can be done:
+
+* The CID can be shared on a broadcast channel using a *Byzantine consistent broadcast* algorithm<sup>[6](#footnotes)</sup>.
+* A new Access node can randomly select a couple of other Access nodes and simply ask them for the CID. If the responses it receives are concordant, it can go ahead and download the data. Once the download is complete, it can generate the state trie for each block and compare its hash to the one found on-chain in the corresponding Execution Result. If it finds an inconsistency, it has all the evidence it needs to submit a slashing challenge against the nodes which provided it with the bad CID.
+
+### Sharding
```
AlexHentschel, line 183:
```suggestion
### Horizontal scaling of historic state storage
```

roadmap/state_sync_protocol/state-sync-protocol.md
```diff
@@ -0,0 +1,204 @@
...
+
+The Execution Data can be sharded so that each Access node only needs to store and respond to requests for a fraction of the entire dataset.
+
+There are many ways that sharding can be implemented. For example, consistent hashing can be used to determine the assignment of Execution Data to Access nodes, so that when nodes join or leave the network at each epoch, the Execution Data can be remapped according to the consistent hashing scheme. We should use the Random Beacon to determine each Access node's locations around the consistent hashing circle in order to prevent malicious Access nodes from purposely skewing the distribution.
+
+It remains a challenge to ensure that Access nodes keep Execution Data available instead of pruning it and ignoring Bitswap requests. *Data availability proofs* and *proofs of retrievability* may be helpful here:
```
AlexHentschel, line 189: Here, I would focus on the ability to horizontally scale the historic state storage within a single access node (instead of sharing across different access nodes)

AlexHentschel (approved):
I'd suggest to re-frame the section about [Future Optimizations](https://github.com/onflow/flow/blob/24cecdcfe0dcb88331356b0e80342e1ce4fa0df7/roadmap/state_sync_protocol/state-sync-protocol.md#future-optimizations) and instead of sharding, talk about horizontal scaling:
* Essentially, the Access nodes become the keeper of the historic state. This state might be valuable for a lot of external clients and the operator of the access node itself. Therefore, I think keeping the _entire_ archival state is a valuable business for some of the access nodes.
* In the [Memory / Storage Estimate](https://github.com/onflow/flow/blob/24cecdcfe0dcb88331356b0e80342e1ce4fa0df7/roadmap/state_sync_protocol/state-sync-protocol.md#memory--storage-estimate) section, you summarize the disk and Ram requirements for keeping state on a single machine. Great. 
   One possibility would be to distribute the state over _different_ access nodes. But then we have to answer the question of how do we ensure that state is sufficiently replicated across many access nodes? In other words, we make the state replication a protocol-layer problem, which is very hard to properly solve.
   In the development of Flow, we assume that the protocol is of wide-spread relevance; hence there is a strong protocol-external incentive for many access nodes to keep a _complete history_ of the execution state (certainly, this is the case for Dapper Labs, and probably many other node operators, exchanges etc). Then, finding nodes who have the full state history is not a problem. Rather, the challenge is for another business entity who also wants to replicate the state history is how to store it within their (single) access node. 
   Instead of talking about sharding (distributing between different nodes), I would prefer if we would focus on horizontal scaling (distributing the history over several machines that all belong to the _same_ access nodes).

AlexHentschel:
Thanks @smnzhu for your patience and the many iterations. A very comprehensive and precise proposal. 👏

PR #692, opened 2021-11-18
Initial commit to capture spork data for Flow into a JSON

franklywatson:
Closes: #???

## Description

This initial commit is the baseline config pulled from https://github.com/onflow/flow/blob/master/docs/content/node-operation/past-sporks.mdx but now in JSON

______


franklywatson:
Will add the fix the main doc to go with this as part of the PR

PR #693, opened 2021-11-18
Initial add of sporks.json file to reflect the current live docs and also fix some errors

franklywatson:
Closes: #???

## Description

The newly added sporks.json is intended for automated use cases and reflect the contents of the current live doc https://docs.onflow.org/node-operation/past-sporks/. This commit also fixes some errors in the current docs


sporks.json
```diff
@@ -0,0 +1,950 @@
+{
```
vishalchangrani, line 1: do we want both then? past-sporks.mdx and this file?

10thfloor: We could use this file to populate the docs pages. That might be handy.

vishalchangrani (approved):

sporks.json
franklywatson: Good question Vishal. I guess it depends if we think it's best to share all the details in both places, or if we're keeping the .mdx file a subset.

Kay-Zee: I like the idea of using the json to generate the mdx file

Kay-Zee (approved):
I noticed there are still a lot of empty space, are there plans to backfill this data? e.g. the cadence tags, and the older access nodes?

franklywatson:
> 

@Kay-Zee yes we should probably backfill the data that is missing, though it could take some time. Perhaps something we can outsource?

PR #696, opened 2021-11-19
modifying collections to only return transaction IDs instead of the f…

vishalchangrani:
…ull transaction; Adding field _self to links

We need some changes to the Swagger definition - 
1. Links object should contain a property of string type called `_self`
2. Removing the following `expand` fields from `Transaction`:
  `proposal_key`
  `authorizers`
 `payload_signatures`
 `envelope_signatures`
Reason: 
- Each of the expand field requires an additional endpoint to be added to the API to support retrieval of the object with a link when the field is not specified in the expand list e.g. `/v1/envelope_signature/<signature>`. This adds a lot of endpoins to the API making it more complicated that it should be.
- There is no savings in terms of lookup for these expand fields and they are not being looked up separately
- Most of the time authorizers are very few in number.

3. Removing the `expand` field `events` from `TransactionResult` for similar reasons as above.

**Reverted changes**:

~~- get collection must return light collection instead of the collection with transactions~~ as explained by @smnzhu, this can be achieved by using the `select=transaction.id` functionality.
~~- execute script must support `latest` as value for block ID or block height~~ added a comment to indicate that if neither block ID nor height is specified, then the script will be executed at the latest sealed block height.

Closes: #???

## Description

______

For contributor use:


openapi/go-client-generated/api/swagger.yaml
```diff
@@ -682,103 +683,16 @@ components:
...
-            result: http://example.com/aeiou
-            envelope_signatures: envelope_signatures
-            payload_signatures: payload_signatures
-            authorizers: authorizers
-            proposal_key: proposal_key
-          proposal_key:
-            sequence_number: 1
-            key_index: 6
+        - null
+        - null
```
sideninja, line 695: Why are these null needed?

sideninja, line 695: Maybe we should define examples, but we can do that after schema is complete

sideninja (approved):
LGTM 🚀

smnzhu:
> get collection must return light collection instead of the collection with transactions

Actually, my intention was that by default it returns the full collection.

If someone wants a light collection, they should use `select` to only select the transaction IDs.

The implementation will need to differentiate between what is requested, but this should be transparent to the client.

smnzhu (changes requested):
> get collection must return light collection instead of the collection with transactions

Actually, my intention was that by default it returns the full collection.

If someone wants a light collection, they should use select to only select the transaction IDs.

The implementation will need to differentiate between what is requested, but this should be transparent to the client.

vishalchangrani:
> > get collection must return light collection instead of the collection with transactions
> 
> Actually, my intention was that by default it returns the full collection.
> 
> If someone wants a light collection, they should use select to only select the transaction IDs.
> 
> The implementation will need to differentiate between what is requested, but this should be transparent to the client.

ic....ok, in that case, I will switch back the spec to return full transaction

openapi/access.yaml
```diff
@@ -331,7 +331,7 @@ paths:
           $ref: '#/components/responses/500InternalServerError'
```
vishalchangrani, line 331: only access.yaml file was changed (the rest of the files is generated code)

smnzhu:
> Each of the expand field requires an additional endpoint to be added to the API to support retrieval of the object with a link when the field is not specified in the expand list e.g. /v1/envelope_signature/<signature>. This adds a lot of endpoins to the API making it more complicated that it should be.

I don't think this is true. Just because a field is expandable, doesn't mean it needs its own endpoint.

smnzhu:
Now, I am not completely against removing some expandable fields, but I want to get rid of the misconception that all expandable fields must have a link. Only fields which are considered "sub-resources" need their own link.

In fact, it is also possible that a non-expandable field (ie a field that is always present) could still have its own link.

If this is the *only* reason for removing these expandable fields, then I don't think it's necessary.

openapi/access.yaml
```diff
@@ -541,14 +541,6 @@ components:
...
           type: object
           properties:
-            proposal_key:
-              type: string
-            authorizers:
-              type: string
-            payload_signatures:
-              type: string
-            envelope_signatures:
-              type: string
```
smnzhu, line 551: If we remove these as expandable fields, we need to add them as `required` above.

smnzhu: Similar for the TransactionResult

vishalchangrani: ahh! good catch.fixing it now

vishalchangrani: done

vishalchangrani:
> > Each of the expand field requires an additional endpoint to be added to the API to support retrieval of the object with a link when the field is not specified in the expand list e.g. /v1/envelope_signature/. This adds a lot of endpoins to the API making it more complicated that it should be.
> 
> I don't think this is true. Just because a field is expandable, doesn't mean it needs its own endpoint.

so my understanding was that an expandable field is something that is being returned as a link so that the client can then choose to retrieve that field. What other values can an exapandable have if not a link? 
If it will be a link but it is not provided as a REST endpoint, then the generated client will not be able to retrieve it as a method call but will have do a regular http GET for the resource and parse the response json. Most users of the API, I am assuming will use a generated client in some language.

vishalchangrani:
> Now, I am not completely against removing some expandable fields, but I want to get rid of the misconception that all expandable fields must have a link. Only fields which are considered "sub-resources" need their own link.
> 
> In fact, it is also possible that a non-expandable field (ie a field that is always present) could still have its own link. Whether a field is expandable has nothing to do with whether it has a link or not.
> 
> If this is the _only_ reason for removing these expandable fields, then I don't think it's necessary.

That is one of the reason, the other reason being that these expandable don't save any processing time and have minimal impact on the payload size.

sideninja:
@smnzhu 
> Now, I am not completely against removing some expandable fields, but I want to get rid of the misconception that all expandable fields must have a link. Only fields which are considered "sub-resources" need their own link.

I agree. Can we assume that only model definitions that have `links` property are supposed to be addressable with their own endpoint?

vishalchangrani:
> @smnzhu
> 
> > Now, I am not completely against removing some expandable fields, but I want to get rid of the misconception that all expandable fields must have a link. Only fields which are considered "sub-resources" need their own link.
> 
> I agree. Can we assume that only model definitions that have `links` property are supposed to be addressable with their own endpoint?

So for example in the following snippet, `payload` is not really a subresource since it does not has an endpoint. However, the client will attempt to retrieve the payload via `/blocks/5422b1a8d5479f82b51d514d451567b5bc0973ad3550f2873c9f44c71231851f/payload`. Shouldn't `/block/{id}/payload` then be added as a new endpoint?

> "_expandable": {
>         "payload": "/blocks/5422b1a8d5479f82b51d514d451567b5bc0973ad3550f2873c9f44c71231851f/payload",
>         "execution_result": "/execution_results/a4eacde933d9dd8b547eeee4ac408eba323f9629c70e1989a74d0d2ace9fa354"
>     },

smnzhu:
> I agree. Can we assume that only model definitions that have links property are supposed to be addressable with their own endpoint?

This is accurate, yes :)

smnzhu (approved):
Looks good

Issue #700, opened 2021-11-24
Change sporks.json structure


contil:
lookig at https://github.com/onflow/flow/blob/139cd856b0eed7ee18763329ec155eeded17bf8e/sporks.json#L2 and https://github.com/onflow/flow/blob/139cd856b0eed7ee18763329ec155eeded17bf8e/sporks.json#L381

I propose to change the structure of the file to add a nesting to something like this :

{
  environments: [
     {      
         env-name: "mainnet"
         iteractions: [
             {
                  iter-name: "14"
.....
            }
     } 
     {      
         env-name: "testnet"
             {
                  iter-name: "31"
.....
            }

Kay-Zee:
I'd probably still keep the "spork name" value of `candidate-9` somewhere in the json.

I'm fine with further nesting though, it might be worth mentioning why you feel the additional nesting is superior.

contil:
I think nesting would help while using tools like jq to query and also to group / import new sporks containing only a subdoc for the latest spork ( i.e. imagine to have a subdoc with only mainnet14 and then use jq / shell to create a full env:"mainnet"["1","2".."14"]

the less we leave full access to the full json the less problems we'll probably have with json-linter, indentation ,convert to yaml etc

Kay-Zee:
any issues from your side @franklywatson ?

franklywatson:
No concerns, but would be useful if we can agree on the element names in scope for this change: 

* iteractions - i assume you mean iterations? @contil 
* iter .. means iteration?
* env-name... maybe we want to speak directly to blockchain domain here and use 'network'?

Want to make sure the final version makes things clear. I think ID could be better than iter/iteration, very direct a clear and backed up by the name. Just wondering if there is a better name for 'iterations' to hold the list

contil:
I agree on the terminology ... I borrowed "environment " and "iteration" ( with a typo ) from the TF code but please use the more appropriate terminology that makes sense ....

franklywatson:
👍🏼  So for final confirmation, how about these names and structure:

```
{
  "networks": [
    {
      "name": "mainnet"
      "sporks": [
        {
          "id": "14"
        }
      ]
    },
    {
      "name": "testnet"
      "sporks": [
        {
          "id": "32"
        }
      ]
    }
  ]
}
```

franklywatson:
[...]

PR #701, opened 2021-11-25
Jerome sporks file

franklywatson:
Closes: #700 

## Description

* Restructure sporks.JSON for better query-ability and extract spork ID into numeric field
* Add missing Cadence tags 
______

For contributor use:

- [ X] Targeted PR against `master` branch
- [ X] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.

### DIFF
--- a/docs/content/node-operation/past-sporks.mdx
+++ b/docs/content/node-operation/past-sporks.mdx
@@ -249,7 +249,7 @@ While FCL v0.0.67 is not strictly necessary to use Mainnet 6, we strongly recomm
 | **Root Parent ID**: 5bc2b0900a5138e39d9209a8fe32e14b3e5c884bd36d2a645620f746b7c8bd47   |
 | **Root State Commit**: 6b9161a225b087a461ec95e710fdf4e73f6d6c9401ebf066207a021dced4ce5e |
 | **Git Commit**:                                                                        |
-| **Branch/Tag**: v0.8.2                                                                 |
+| **Branch/Tag**:                                                                  |
 | **Docker Tag**:                                                                        |
 
 ---
@@ -309,7 +309,7 @@ While FCL v0.0.67 is not strictly necessary to use Mainnet 6, we strongly recomm
 | **Root Parent ID**: 68c2bbe68524b50f5d689bc2ac7ad2dd70e88ed7dd15ad6c3cdf6ea314cb1aa3   |
 | **Root State Commit**: c05086e4d1d428d3b9af5bd8b81d8780054f783ef4eec3ca28b491202e9ac696|
 | **Git Commit**: b9b197280d6590576f1ef183bc3d04d41d6be587                               |
-| **Branch/Tag**: v0.9.3                                                                 |
+| **Branch/Tag**:                                                                  |
 | **Docker Tag**:                                                                        |
 
 ---
--- a/sporks.json
+++ b/sporks.json
@@ -25,7 +25,7 @@
       "flow-dps-tag": "0.1",
       "cadence-tag": "",
       "docker-tag": "v0.22.9-patch-1-epoch-view-check-hotfix",
-      "access-node": "access.mainnet.nodes.onflow.org:9000"
+      "access-node": "access-001.mainnet14.nodes.onflow.org:9000"
     },
     {
       "spork-id": "Mainnet-13",
--- a/sporks.json
+++ b/sporks.json
@@ -1,952 +1,995 @@
 {
-  "mainnet": [
+  "networks" : [
     {
-      "spork-id": "Mainnet-14",
-      "spork-date": "2021-10-06T16:00:00Z",
-      "root-height": "19050753",
-      "root-parent-id": "ac4dbf344ce96e39e15081f1dc3fbbf6dc80532e402de9a57af847d3b35df596",
-      "root-state-commit": "641eb088e3ce1a01ff56df2d3a14372c65a7fef44c08799eb92cd7759d1d1d2a",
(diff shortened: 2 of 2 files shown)


contil (approved):
LGTM

docs/content/node-operation/past-sporks.mdx
```diff
@@ -249,7 +249,7 @@ While FCL v0.0.67 is not strictly necessary to use Mainnet 6, we strongly recomm
 | **Root Parent ID**: 5bc2b0900a5138e39d9209a8fe32e14b3e5c884bd36d2a645620f746b7c8bd47   |
 | **Root State Commit**: 6b9161a225b087a461ec95e710fdf4e73f6d6c9401ebf066207a021dced4ce5e |
 | **Git Commit**:                                                                        |
-| **Branch/Tag**: v0.8.2                                                                 |
```
Kay-Zee, line 252: why were these removed?

sporks.json
```diff
@@ -1,950 +1,995 @@
...
+      "sporks": [
+        {
+          "id": 14,
+          "name": "Mainnet-14",
+          "spork-date": "2021-10-06T16:00:00Z",
+          "root-height": "19050753",
+          "root-parent-id": "ac4dbf344ce96e39e15081f1dc3fbbf6dc80532e402de9a57af847d3b35df596",
+          "root-state-commit": "641eb088e3ce1a01ff56df2d3a14372c65a7fef44c08799eb92cd7759d1d1d2a",
+          "git-commit-hash": "f019c1dbd778ce9f92dea61349ca36003678a9ad",
+          "state-artefacts" : {
```
Kay-Zee, line 14: interesting that we decided to go with this spelling of artefact

sporks.json
```diff
@@ -1,950 +1,995 @@
...
-      "docker-tag": "v0.12.1",
-      "access-node": ""
+  "networks" : [
+    {
+      "name": "mainnet",
+      "sporks": [
+        {
+          "id": 14,
+          "name": "Mainnet-14",
+          "spork-date": "2021-10-06T16:00:00Z",
```
Kay-Zee, line 9: this includes the time, so maybe spork-time?

sporks.json
```diff
@@ -1,950 +1,995 @@
...
+              "execution-state-wal-chunks": ""
+            }
+          },
+          "flow-go-tag": "v0.10.1",
+          "flow-dps-tag": "",
+          "cadence-tag": "v0.9.2",
+          "docker-tag": "v0.10.1",
+          "access-node": "access-001.mainnet1.nodes.onflow.org:9000"
+        },
+        {
```
Kay-Zee, line 399: haha, that's interesting interpretation, 0.9.

sporks.json
```diff
@@ -1,950 +1,995 @@
...
+          "id": 0.9,
+          "name": "Candidate-9",
+          "spork-date": "2020-09-25T16:00:00Z",
+          "root-height": "6483246",
+          "root-parent-id": "9131733835702b0d6321088bddb4642a4964bb5c630440ccb0de47bdbc371d1a",
+          "root-state-commit": "90c6f406f5d21880d525ad4702cb249509b85e7f745db2de67e9fe541a56da4c",
+          "git-commit-hash": "",
```
Kay-Zee, line 406: any reason these are empty? just not that important since they're candidates?

sporks.json
```diff
@@ -1,950 +1,995 @@
...
+              "execution-state-wal-chunks": ""
+            },
+            "s3": {
+              "root-checkpoint-file": "",
+              "root-protocol-state-snapshot": "",
+              "node-info": "",
+              "execution-state-wal-chunks": ""
+            }
+          },
+          "flow-go-tag": "",
```
Kay-Zee, line 449: and these as well

Kay-Zee (approved):

docs/content/node-operation/past-sporks.mdx
franklywatson: I removed it because I couldn't find either a branch or a tag that has this version.

sporks.json
franklywatson: Yeah this is the spork time, but adjusted for UTC given the time format

sporks.json
franklywatson: Yeah, if we went with full numerics in the candidates then the other full Mainnet IDs would duplicate these and so numeric searches would have issues

sporks.json
franklywatson: I couldn't find it in the Sporks history notion, and also there is no git commit hash for this version in the current public docs

sporks.json
franklywatson: Same, again - there's nothing noted in the Sporks notion doc and nor in the current public docs for past sporks. Any suggestions you have to identify the missing data would be great

sporks.json
franklywatson: Happy to change it LMK

sporks.json
Kay-Zee: I meant the naming of the field

sporks.json
Kay-Zee: for this particular field, since we have the git tag, we can get the commit hash if we just look at the commit that's tagged.

sporks.json
franklywatson: Ah, fair point. Will do

sporks.json
franklywatson: Ok thanks will do, my git-fu is a little rusty still

PR #703, opened 2021-11-29
add FLIP for cadence mutability restrictions

dsainati1:
Adds FLIP for improvements to Cadence's handling of public fields in structures. See discussion in https://github.com/onflow/cadence/issues/1260 for more discussion of this issue. An implementation of the FLIP can be found in https://github.com/onflow/cadence/pull/1267.

______

For contributor use:


flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [703](https://github.com/onflow/flow/pull/703)      |
```

bluesign:
First of all, thank you very much for taking time for the detailed FLIP.

I have a question about: 

> Another approach would be to make `let` a true constant declaration, and forbid writing or mutating it in any context,
>  while allowing `var` to be mutated in all the contexts it can be written. This, however, is likely too restrictive;
>  a complex initialization for a dictionary or array value may benefit from the ability to iteratively mutate its contents,
>  up to the point at which its value is finalized, and can thus be exported. As such we would like to still allow users
>  to perform mutations in an limited context.

This part confused me a little, I see following scenarios in general: ( if we try to limit to pub scope)

owner can change, public can change: `pub(set) var`
owner can change, public cannot: `pub var`
owner and public cannot change: `pub let`

if I would like to let's say iteratively change the content, it feels like I should choose `pub var`

dsainati1:
This is a good question, and the semantics you propose would also solve the issue in question. Coming from a functional background myself, I actually personally prefer the behavior you describe in general. However, given the existing programming paradigm used by Cadence, I am not sure that having `pub let` denote a purely immutable value by default is the right solution. There is a pretty big difference between overwriting a value entirely (like `var` lets you do), and adding or removing a value from an array or dictionary. Requiring that users make their fields writable in order to mutate them does not seem like the right approach to me for this reason. 

Additionally, there are existing contracts that we have to consider, and it is likely that restricting `pub let` fields from being mutated in any context, even inside the contract or structure where they are defined, would break a much larger proportion of existing code. Mutating a field privately is not inherently a bad behavior, and is thus much more likely to occur in valid code than the field being mutated in a public context, which is a security issue. 

This FLIP proposes a minimally invasive solution to the existing problem in the interest of presenting an easy migration path for users who were relying on the old, bad behavior by only restricting exactly the bad behavior we wish to prevent. That said, if the community overall prefers the behavior you are describing, it is certainly also worth considering since it does address the problem at hand.

bluesign:
Considering Cadence dictionaries are value types, I think `let + value type` to be mutable will be little unexpected.   

But rest of the part I agree, it is a bit bigger change for backward compatibility. (I just checked contracts, lots of `let`)

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+        }
+    }
+
+    init() {
+        self.x = S()
+    }
+}
+
+pub fun bar() {
+    Foo.x.y.remove(at: 0) // cannot mutate `y`, field was defined in `S`
```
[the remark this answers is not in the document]
dsainati1: here `Foo` is a contract, which does not need to be instantiated like a struct or resource would be

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+Specifically, the error is emitted whenever a user attempts to perform an 
+index assignment on a member that is not either declared with the `pub(set)` 
+access mode, or is defined in the current enclosing scope. This check is the
+same one performed for writing to fields, with the difference that mutation 
+is allowed on both `let` and `var` fields, while only the latter can be written to.
+
+Additionally, array and dictionary methods now track an additional bit of information
+indicating whether they mutate their receiver. Mutating methods may not be called on 
+members that are not declared with the `pub(set)` access mode, or defined in the current
+enclosing scope.
```
turbolent, line 140: Would we want to expose this to Cadence programs? Swift has the `mutating` function modifier for this purpose.

Just curious, I don't think that is necessary for this proposal, it could be added through a future proposal.

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    pub fun getFoo(): Foo {
+        return self.foo
+    }
+}
+
+pub fun main() {
+    let a = C.getFoo()
+    a.arr.append(0) // a.arr is now [3, 0]
```
turbolent, line 220: Is this what is proposed here? Currently, this should not be expected, as the field's value is copied

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+viewed. Within a struct, a `pub let` array field would have the type `[Int]`, permitting all array 
+operations on it, including mutating ones. Outside the struct, the same field might have the type 
+`[Int]{ReadonlyArray<Int>}`, in effect a restricted type that only permits the non-mutating 
+operations to be performed on that array. This, however, is overly complex, and introduces
+a new type to accomplish the same purpose as what this FLIP proposes. 
+
+### Performance Implications
+
+* This is a change only to the Cadence type checker, so it will not have any 
+performance implications for Flow or the Cadence runtime.
```
turbolent, line 234: As type checking is currently run as part of execution, it implicitly has an effect on total performance of the system

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+is allowed on both `let` and `var` fields, while only the latter can be written to.
+
+Additionally, array and dictionary methods now track an additional bit of information
+indicating whether they mutate their receiver. Mutating methods may not be called on 
+members that are not declared with the `pub(set)` access mode, or defined in the current
+enclosing scope. 
+
+The array methods that are considered mutating are `append`, `appendAll`, `remove`,
+`insert`, `removeFirst` and `removeLast`. The dictionary methods that are considered
+mutating are `insert` and `remove`.
```
turbolent, line 144: Will only these methods, or also the operators performing these operations be considered mutating? For example, `let xs: {"a": 1}; xs["a"] = 2`. I think they should and probably are implicit here, but maybe mention it explicitly

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+allows any code to mutate or write to the field it applies to. 
+
+### Implementation Details 
+
+This change adds a new error, the `ExternalMutationError`, which is raised when a field 
+is mutated outside of the context in which it was defined. The error message will also
+suggest that the user instead use a setter or modifier method for that field.
+
+Specifically, the error is emitted whenever a user attempts to perform an 
+index assignment on a member that is not either declared with the `pub(set)`
```
turbolent, line 132: Does the proposal specifically only handle arrays and dictionaries, and the indexing assignment, or does it also address field assignments on composite values?
For example:

```rust
pub struct Foo {
  pub var name: String
  init() {
    self.name = "initial"
  }
}

pub struct Bar {
  pub let foo: Foo
  init() {
    self.foo = Foo()
  }
}

let bar = Bar()
bar.foo.name = "new"  // is this also rejected?
```

turbolent (commented):
Looks great!

flips/20211129-cadence-mutability-restrictions.md
dsainati1: When I run this code as a contract/transaction pair on the playground, the output is `[3,0]`. Is that not expected?

```
pub contract C {
    pub struct Foo {
        pub let arr : [Int]
        init() {
            self.arr = [3]
        }
    }
    
    priv let foo : Foo
    init() {
        self.foo = Foo()
    }
    pub fun getFoo(): Foo {
        return self.foo
    }
}
```
and 
```
import C from 0x01

transaction {

  prepare(acct: AuthAccount) {}

  execute {
    let a = C.getFoo()
    a.arr.append(0) // a.arr is now [3, 0]
    log(a.arr)
  }
}
```

turbolent: Ah, my bad, I misread the example as the array being returned from the getter and being modified!

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
...
+    pub let y : [Int]
+
+    init() {
+        self.y = [3];
+    }
+}
+
+pub fun addToY(foo: Foo, i: Int) {
+    foo.y.append(i)
+}
```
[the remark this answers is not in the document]
dsainati1: Yeah I describe this below, but the builtin array/dictionary functions are given a bit tracking whether they mutate the receiver or not.

flips/20211129-cadence-mutability-restrictions.md
dsainati1: We could; unsure what the use case is for this, but it's possible.

flips/20211129-cadence-mutability-restrictions.md
dsainati1: Yes indexed assignment is also considered mutating.

bluesign:
I think it would be also very nice if we can get immutability as a concept in cadence. 

It can save a lot of code in contracts (especially query part of some contracts) 

I am thinking something like a flag in composite or resource, that will make the composite reference immutable. 

Basically you can access all fields and call all methods as long as it doesn't mutate the object. (Ofc you cannot move resources)

flips/20211129-cadence-mutability-restrictions.md
bluesign: I think this is the case : 
`pub struct Foo {
  pub(set) var name: String
  init() {
    self.name = "initial"
  }
}

pub struct Bar {
  pub let foo: Foo
  init() {
    self.foo = Foo()
  }
}

let bar = Bar()
bar.foo.name = "new"  // is this also rejected?`

turbolent: @dsainati1 Sorry, I wrote the code in the comment field and missed the essential part, like @bluesign corrected, `Foo.name` should be `pub(set)`. 

The point I'm trying to make here: Just like the proposal is currently proposing to improve working with containers such as arrays and dictionaries, which are mutable types, composites can also be mutable.

dsainati1: Yea this is an interesting point, and one where I think there is room for more discussion on the design. It is certainly possible that we should restrict writes to composites like this: i.e. if any of the members in an `x.y.z = e` chain are declared with something either than `pub(set) var`, we might not allow them to be mutated. However, given that `pub(set) var`s must be explicitly declared by users, I don't believe they as likely to be accidentally mutated. 

In my opinion, I think we can use the same behavior for mutations we currently use for writes, wherein a `pub(set)` field can still be written to even if it is a member of a composite that is stored in a `let` field. However, I am curious what arguments there are for a different approach.

joshuahannan: Would this apply to calling functions on composite fields also? I don't think it would, but maybe something worth considering

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+This has the potential to break a number of existing contracts by restricting 
+code that was previously legal. This change would require a migration path for 
+these contracts in order for developers to update their contracts to satisfy
+Cadence's new restrictions.
+
+This also removes a small amount of expressivity from the language. Previously, 
+the `pub let` declaration was a way to create a field that could be read and
+mutated in all contexts, but not written, while `pub(set) var` created
+a field that could be read, written and mutated in all contexts. After this change,
+it would not be possible to describe a field that can be read and mutated but not set,
```
joshuahannan, line 164: Are there cases where this is needed that a setter function would be inadequate?

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+    foo.x.insert(key: 0, 3) // ok, pub(set) access modifier allows mutation
+}
+```
+
+### Compatibility
+
+This change is not backwards compatible with existing Cadence smart contracts, but
+is being made in service of a future version of Cadence wherein no backwards incompatible
+changes will be required going forwards. It should not affect any other parts of Flow, 
+however.
```
joshuahannan, line 361:
```suggestion
```
However probably isn't necessary?

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+pub struct Bar {
+  pub(set) var content: String
+  init() {
+    self.content = "original"
+  }
+}
+```
+
+This is one potential way to get around mutability restrictions on `pub let` fields, by aliasing such
+a field with a `pub(set) var`. Is preventing this kind of issue within the scope of this change? How
```
joshuahannan, line 407: This is analogous to a setter function to me, so it seems fine to allow this

joshuahannan:
This is exciting! I'm looking forward to having this in the language. 😃

flips/20211129-cadence-mutability-restrictions.md
dsainati1: IMO, no, there are no specific cases where you could not accomplish the same thing with a setter function, but I felt it was worth at least mentioning, since it is a concept that can no longer be expressed.

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,417 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [703](https://github.com/onflow/flow/pull/703)       |
+| **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Updated**   | 2021-11-29                                           |
```
dsainati1, line 8:
```suggestion
| **Updated**   | 2021-12-08                                           |
```

j1010001 (commented):
Just some minor typos and style changes.

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+correctness of contracts deployed to Flow. 
+
+## Design Proposal
+
+### Explanation 
+
+Cadence controls where and to what extent variables can be read and written using a combination of
+access modifiers and declaration kinds, as described [here](https://docs.onflow.org/cadence/language/access-control/).
+Of note is that the `let` kind does not allow fields to be written to in any scope, whereas `var` allows them
+to be written in the "Current and Inner" scopes; that is, the scope in which the field was declared, and any scopes
```
turbolent, line 61: "written to" sounds like mutation is not allowed, but in fact it is, and this is exactly what this proposal is trying to introduce.

Maybe reword to something like "reassigned", etc.

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+
+    init() {
+        self.x = [3]
+    }
+}
+
+pub fun bar() {
+    let foo = Foo()
+    foo.x = [0] // writes to x, not allowed
+    foo.x[0] = 0 // does not write to x, currently allowed
```
turbolent, line 78:
```suggestion
    foo.x = [0] // reassignment to `x` is not allowed, as variable is `let`
    foo.x[0] = 0 // mutates the array, but does not reassign to `x`, currently allowed
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+in the current and inner scopes, the same contexts in which the field could be written to if it were a `var`. So the following 
+would typecheck:
+
+```
+pub struct Foo {
+    pub let x : [Int]
+
+    init() {
+        self.x = [3];
+    }
```
turbolent, line 94:
```suggestion
```cadence
pub struct Foo {
    pub let x: [Int]

    init() {
        self.x = [3]
    }
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+```
+
+while the following would not:
+
+```
+pub struct Foo {
+    pub let y : [Int]
+
+    init() {
+        self.y = [3];
```
turbolent, line 109:
```suggestion
```cadence
pub struct Foo {
    pub let y: [Int]

    init() {
        self.y = [3]
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+    }
+}
+
+pub fun addToY(foo: Foo, i: Int) {
+    foo.y.append(i)
+}
```
turbolent, line 115: Maybe demonstrate using a "script function" and add an explanation comment:

```suggestion
pub fun main() {
    let foo = Foo()
    foo.y.append(1) // invalid, field `y` is let and cannot be mutated
}
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+```
+pub struct Foo {
+    pub let x : [Int]
+
+    init() {
+        self.x = [3];
+    }
+
+    pub fun addToX(i: Int) {
+        self.x.append(i)
```
turbolent, line 97:
```suggestion
        self.x.append(i)  // mutation of field `x` in same scope is valid, even though field is `let`
```

flips/20211129-cadence-mutability-restrictions.md
turbolent: I like the proposal, but it seems like it only addresses part of the general problem, i.e. only focuses on arrays and dictionaries, but composite values are also containers. 

Then again, maybe it is good to keep the scope of this proposal narrow, and we can extend the new behaviour to composites in a follow-up proposal.

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+```
+pub contract C {
+    pub struct Foo {
+        pub let arr : [Int]
+        init() {
+            self.arr = [3]
+        }
+    }
+    
+    priv let foo : Foo
```
turbolent, line 208:
```suggestion
```cadence
pub contract C {
    pub struct Foo {
        pub let arr: [Int]
        init() {
            self.arr = [3]
        }
    }
    
    priv let foo: Foo
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+
+Additionally, array and dictionary methods now track an additional bit of information
+indicating whether they mutate their receiver. Mutating methods may not be called on 
+members that are not declared with the `pub(set)` access mode, or defined in the current
+enclosing scope. 
+
+The array methods that are considered mutating are `append`, `appendAll`, `remove`,
+`insert`, `removeFirst` and `removeLast`. The dictionary methods that are considered
+mutating are `insert` and `remove`. Additionally, indexed assignment (`x[y] = e`) will
+be considered mutating for both arrays and dictionaries.
```
turbolent, line 145:
```suggestion
mutating are `insert` and `remove`. Additionally, indexed assignment (e.g. `x[y] = e`) will
be considered mutating for both arrays and dictionaries.  
```

Maybe mention that this will apply to all assignment operators, e.g. also the swap operator (`<->`)

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+### Examples
+
+Some examples of code that produces a type error as a result of this restriction:
+
+```
+pub resource Foo {
+    pub let x : {Int: Int}
+
+    init() {
+        self.x = {0:3};
```
turbolent, line 255:
```suggestion
```cadence
pub resource Foo {
    pub let x: {Int: Int}

    init() {
        self.x = {0: 3}
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+```
+pub struct Bar {
+    pub let foo: Foo
+    init() {
+        self.foo = Foo()
+    }
+}
+
+pub struct Foo {
+    pub let x : [Int]
```
turbolent, line 275:
```suggestion
    pub let x: [Int]
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+```
+
+```
+pub contract Foo {
+    pub let x : S
+    
+    pub struct S {
+        pub let y : [Int]
+        init() {
+            self.y = [3]
```
turbolent, line 295:
```suggestion
```cadence
pub contract Foo {
    pub let x: S
    
    pub struct S {
        pub let y: [Int]
        init() {
            self.y = [3]
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+pub fun bar() {
+    Foo.x.y.remove(at: 0) // cannot mutate `y`, field was defined in `S`
+}
+```
+```
+pub contract Foo {
+    pub let x : S
+    
+    pub struct S {
+        pub let y : [Int]
```
turbolent, line 313:
```suggestion

```cadence
pub contract Foo {
    pub let x: S
    
    pub struct S {
        pub let y: [Int]
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+```
+
+while the following are allowed:
+
+```
+pub struct Foo {
+    pub let x: {Int: Int}
+
+    init() {
+        self.x = {3:3};
```
turbolent, line 333:
```suggestion
```cadence
pub struct Foo {
    pub let x: {Int: Int}

    init() {
        self.x = {3: 3}
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+        foo.x[0] = 3 // ok, mutation occurs inside defining struct Foo
+    }
+}
+```
+```
+pub struct Foo {
+    pub(set) var x: {Int: Int}
+
+    init() {
+        self.x = {3:3};
```
turbolent, line 347:
```suggestion

```cadence
pub struct Foo {
    pub(set) var x: {Int: Int}

    init() {
        self.x = {3: 3}
```

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,418 @@
...
+
+This will change the way users write their smart contracts. Given that this is preventing
+users from writing an unsafe pattern, we do not expect this to invalidate a particularly
+large amount of existing code. However, any code that is invalidated by this change can be
+refactored to expose a setter/modifier method for affected fields. 
+
+## Related Issues
+
+Do we wish to handle aliased references? Consider the following example (courtesy of @SupunS):
+```
```
turbolent, line 373:
```suggestion
```cadence
```

turbolent (approved):
+1 from me, this proposed change will improve safety in contracts.

As mentioned above, I feel like the "`pub let` mutability problem" won't be solved completely after this change, so it would be nice to also think about how these proposed mutation rules can be applied to composite values.

dsainati1:
> As mentioned above, I feel like the "pub let mutability problem" won't be solved completely after this change, so it would be nice to also think about how these proposed mutation rules can be applied to composite values.

This proposal uses the same behavior that Cadence currently has for reassignments, which is that the access modifier of the "lowest" field in a nested composite determines whether or not something can be reassigned. Consider:

```
pub struct Foo {
    pub let x: Integer
    pub(set) var y: Integer
    init() {
        self.x = 3
        self.y = 3
    }
}

pub struct Bar {
    pub(set) var f: Foo
    pub let g: Foo
    init() {
        self.f = Foo()
        self.g = Foo()
    }
}

pub fun main() {
    let b = Bar()
    b.f.x = 2 // this is an error
    b.g.y = 1 // this is fine
}
```

One possible change that might address your concern would be to consider the access modifiers of every field in the nested access; i.e. require that in order to modify `a.b.c`, both `b` and `c` have a `pub(set) var` modifier. In this scenario, `b.f.x` and `b.g.y` would both not be writeable, since one of the two fields in the access has a `pub let` modifier, but `b.f.y` would be valid to write to from outside of both `Bar` and `Foo`.

joshuahannan (approved):
Still looks good to me!

flips/20211129-cadence-mutability-restrictions.md
dete: I'm really on the fence here, and the example in this document really makes it a tough call.

```
    foo.y.append(1) // invalid, field `y` is let and cannot be mutated
```

In this case, I'm not assigning to `y`, I'm calling a method on `y`. If `y` were a user-defined type with a method called `append()`, this would NOT be an error. To make it an error just because `y` is an language-defined collection type is pretty inconsistent.

OTOH- Containers are the place where this kind of "gotcha" is most likely to be a problem, so maybe it's worth introducing an inconsistency temporarily until we can find a better solution?

I don't feel confident enough to push this discussion in either direction, to be honest.

dsainati1: Looking at this and @turbolent's comments below, the consensus seems to be that this FLIP does not address 100% of the footgun problem, but does hit the most common cases. If we want to handle the parts of this "gotcha" case, as I see it we have a couple options:

1) We can update the FLIP to handle the nested-composite mutation case. There are a couple ways I can see to to do this: we add a `mutating` keyword (or similar) to method definitions, and prevent such methods from being called on fields declared with a `let`. We would also have the option of enforcing that `mutating` methods are the only way to modify internal state (like Swift does), or we could choose to allow the modifier to be used at the developer's discretion, which is easier but also less accurate. 

2) We merge the FLIP now and attempt to address the nested-composite case later. The difficulty here is that any later solution to the nested-composite case would have to be a non-breaking change. Adding a `mutating` keyword as described above, but only restricting where `mutating` methods are used rather than also requiring certain methods to be marked as `mutating` is one way to address this without a breaking change, but it certainly isn't a perfect solution.

3) We merge the FLIP as-is, and just live with the inconsistency/incompleteness. 

My personal preference is for option 2, as long as we can come up with a non-breaking solution that we can apply later that is also a satisfying way to solve the problem. Interested to see if anybody has any thoughts on non-breaking fixes for this.

joshuahannan: I'm fine with just not allowing anything on the array and dictionary fields, since like Dete said, that is the place where the vast majority of people make their mistakes. The inconsistency doesn't bother me much.

bjartek:
+1. It is too easy to by accident make something mutable.

flips/20211129-cadence-mutability-restrictions.md
```diff
@@ -0,0 +1,421 @@
+Cadence Mutability Restrictions
+
+| Status        | (Proposed)       |
```
dsainati1, line 3:
```suggestion
| Status        | (Approved)       |
```

PR #705, opened 2021-12-01
Vishal/more openapi changes

vishalchangrani:
1. Removing `height` as require for the Get Block by height call since height can be absent if `start_height` and `end_height `query params are specified. 2374431dace71bd22901b6d7becd9a14375ef26e
2. The `select` query param lets you unselect any field in the response hence removing all the `require` for elements in the response object. 9a99bbf
3. For the GetCollection by ID call, it will be expensive to look up all the transactions for the collection and the response could be huge. So made a change to make the transaction in the response as `expandable`. If the query param specifies `?expand=transaction` only then will the transactions be looked up and populated in the response, otherwise the `_expandable` field will list all the links (e.g. /v1/transactions/1234556999999) in an array thus skipping the extra lookups for transaction. 93d7a6d

Once y'all approve the PR, I will also include the generated files :) 

## Description

______

For contributor use:

### DIFF
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -479,11 +479,11 @@ components:
           items:
             $ref: '#/components/schemas/Transaction'
         _expandable:
-          type: array
-          items:
-            type: object
-            properties:
-              transaction:
+          type: object
+          properties:
+            transactions:
+              type: array
+              items:
                 type: string
                 format: uri
         _links:
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -27,7 +27,6 @@ paths:
             uniqueItems: true
           explode: false
           style: form
-          required: true
         - name: start_height
           in: query
           schema:
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -478,6 +478,14 @@ components:
           type: array
           items:
             $ref: '#/components/schemas/Transaction'
+        _expandable:
+          type: array
+          items:
+            type: object
+            properties:
+              transaction:
+                type: string
+                format: uri
         _links:
           $ref: '#/components/schemas/Links'
     Transaction:
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -408,9 +408,6 @@ components:
   schemas:
     Account:
       type: object
-      required:
-        - address
-        - balance
       properties:
         address:
           $ref: '#/components/schemas/Address'
@@ -439,14 +436,6 @@ components:
           $ref: '#/components/schemas/Links'
     AccountPublicKey:
       type: object
-      required:
-        - index
-        - public_key
-        - signing_algorithm
-        - hashing_algorithm
-        - sequence_number
-        - weight
-        - revoked
       properties:
         index:
           type: integer
@@ -482,9 +471,6 @@ components:
         - KMAC128
     Collection:
       type: object
-      required:
-        - id
-        - transactions
       properties:
         id:
           $ref: '#/components/schemas/Identifier'
@@ -496,17 +482,6 @@ components:
           $ref: '#/components/schemas/Links'
     Transaction:
       type: object
-      required:
-        - id
-        - script
-        - arguments
-        - reference_block_id
-        - gas_limit
-        - payer
-        - proposal_key
-        - authorizers
-        - payload_signatures
-        - envelope_signatures
       properties:
         id:
           $ref: '#/components/schemas/Identifier'
@@ -551,10 +526,6 @@ components:
           $ref: '#/components/schemas/Links'
     ProposalKey:
       type: object
-      required:
-        - address
-        - key_index
-        - sequence_number
       properties:
         address:
           $ref: '#/components/schemas/Address'
@@ -566,11 +537,6 @@ components:
           format: uint64
     TransactionSignature:
       type: object
-      required:
-        - address
-        - signer_index
-        - key_index
-        - signature
       properties:
         address:
           $ref: '#/components/schemas/Address'
@@ -584,12 +550,6 @@ components:
           $ref: '#/components/schemas/Signature'
     TransactionResult:
       type: object
-      required:
-        - block_id
-        - status
-        - error_message
-        - computation_used
-        - events
       properties:
         block_id:
           $ref: '#/components/schemas/Identifier'
@@ -616,8 +576,6 @@ components:
        - Expired
     Block:
       type: object
-      required:
-        - header
       properties:
         header:
           $ref: '#/components/schemas/BlockHeader'
@@ -637,12 +595,6 @@ components:
           $ref: '#/components/schemas/Links'
     BlockHeader:
       type: object
-      required:
-        - id
-        - parent_id
-        - height
-        - timestamp
-        - parent_voter_signature
       properties:
         id:
           $ref: '#/components/schemas/Identifier'
@@ -658,9 +610,6 @@ components:
           $ref: '#/components/schemas/Signature'
     BlockPayload:
       type: object
-      required:
-        - collection_guarantees
-        - block_seals
       properties:
         collection_guarantees:
           type: array
@@ -674,10 +623,6 @@ components:
           uniqueItems: true
     CollectionGuarantee:
       type: object
-      required:
-        - collection_id
-        - signer_ids
-        - signature
       properties:
         collection_id:
           $ref: '#/components/schemas/Identifier'
@@ -691,11 +636,6 @@ components:
           $ref: '#/components/schemas/Signature'
     BlockSeal:
       type: object
-      required:
-        - block_id
-        - result_id
-        - final_state
-        - aggregated_approval_signatures
       properties:
         block_id:
           $ref: '#/components/schemas/Identifier'
@@ -716,9 +656,6 @@ components:
       pattern: '[a-fA-F0-9]{64}'
     AggregatedSignature:
       type: object
-      required:
-        - verifier_signatures
-        - signer_ids
       properties:
         verifier_signatures:
           type: array
@@ -734,10 +671,6 @@ components:
           uniqueItems: true
     ExecutionResult:
       type: object
-      required:
-        - id
-        - block_id
-        - events
       properties:
         id:
           $ref: '#/components/schemas/Identifier'
@@ -751,12 +684,6 @@ components:
           $ref: '#/components/schemas/Links'
     Event:
       type: object
-      required:
-        - type
-        - transaction_id
-        - transaction_index
-        - event_index
-        - payload
       properties:
         type:
           $ref: '#/components/schemas/EventType'


openapi/access.yaml
```diff
@@ -483,31 +471,25 @@ components:
...
-        - id
-        - transactions
       properties:
         id:
           $ref: '#/components/schemas/Identifier'
         transactions:
           type: array
           items:
             $ref: '#/components/schemas/Transaction'
+        _expandable:
```
vishalchangrani, line 481: Without `expand=transaction`

> {
>   "id": "string",
>   "_expandable": [
>     {
>       "transaction": "/v1/transactions/1234123124"
>     },
>     {
>       "transaction": "/v1/transactions/1234123125"
>     },
>     {
>       "transaction": "/v1/transactions/1234123126"
>     },
>   ],
>   "_links": {
>     "_self": "string"
>   }
> }

vishalchangrani: With `expand=transaction`

> {
>   "id": "string",
>   "transactions": [
>     {
>       "id": "string",
>       "script": "string",
>       "arguments": [
>         "string"
>       ],
>       "reference_block_id": "string",
>       "gas_limit": 0,
>       "payer": "string",
>       "proposal_key": {
>         "address": "string",
>         "key_index": 0,
>         "sequence_number": 0
>       },
>       "authorizers": [
>         "string"
>       ],
>       "payload_signatures": [
>         {
>           "address": "string",
>           "signer_index": 0,
>           "key_index": 0,
>           "signature": "string"
>         }
>       ],
>       "envelope_signatures": [
>         {
>           "address": "string",
>           "signer_index": 0,
>           "key_index": 0,
>           "signature": "string"
>         }
>       ],
>       "result": {
>         "block_id": "string",
>         "status": "Pending",
>         "error_message": "string",
>         "computation_used": 0,
>         "events": [
>           {
>             "type": "string",
>             "transaction_id": "string",
>             "transaction_index": 0,
>             "event_index": 0,
>             "payload": "string"
>           }
>         ],
>         "_links": {
>           "_self": "string"
>         }
>       },
>       "_expandable": {
>         "result": "string"
>       },
>       "_links": {
>         "_self": "string"
>       }
>     },
>     {
>       "id": "string",
>       "script": "string",
>       "arguments": [
>         "string"
>       ],
>       "reference_block_id": "string",
>       "gas_limit": 0,
>       "payer": "string",
>       "proposal_key": {
>         "address": "string",
>         "key_index": 0,
>         "sequence_number": 0
>       },
>       "authorizers": [
>         "string"
>       ],
>       "payload_signatures": [
>         {
>           "address": "string",
>           "signer_index": 0,
>           "key_index": 0,
>           "signature": "string"
>         }
>       ],
>       "envelope_signatures": [
>         {
>           "address": "string",
>           "signer_index": 0,
>           "key_index": 0,
>           "signature": "string"
>         }
>       ],
>       "result": {
>         "block_id": "string",
>         "status": "Pending",
>         "error_message": "string",
>         "computation_used": 0,
>         "events": [
>           {
>             "type": "string",
>             "transaction_id": "string",
>             "transaction_index": 0,
>             "event_index": 0,
>             "payload": "string"
>           }
>         ],
>         "_links": {
>           "_self": "string"
>         }
>       },
>       "_expandable": {
>         "result": "string"
>       },
>       "_links": {
>         "_self": "string"
>       }
>     }
>   ],
>   "_links": {
>     "_self": "string"
>   }
> }

sideninja (approved):
LGTM

sideninja:
@vishalchangrani are we making the change to include `transaction` expandable on collection as part of this PR or another?

sideninja:
Another comment I have is that we need to keep in mind removing required from response fields makes them have `omitempty` json tag, so we just need to be mindful if there are any values that should be present even if empty. 
I personally don't think there is a problem with that (maybe the only one could be empty args on tx but not even that). However this will work nicely with collection being extended with transactions field, since if now we add transaction response with only IDs on collection it will show empty fields too which is not correct.

smnzhu:
@vishalchangrani I'm not so sure about the change to collections. See the example I originally made on how this should be done: https://www.notion.so/dapperlabs/REST-API-Design-9594c690c8a342e383e1d372b5860116#ff2ec5c8469748ac859b60d195ae23ac

vishalchangrani:
> @vishalchangrani are we making the change to include `transaction` expandable on collection as part of this PR or another?

It is part of this PR - the third commit.

vishalchangrani:
> @vishalchangrani I'm not so sure about the change to collections. See the example I originally made on how this should be done: https://www.notion.so/dapperlabs/REST-API-Design-9594c690c8a342e383e1d372b5860116#ff2ec5c8469748ac859b60d195ae23ac

So in that example (copied below), the transactions are looked up by default (if no expandable is specified). That is something I am trying to avoid since it would have a performance hit since each of these transactions are individually looked by by their ID. Hence the proposed change, where by default only the trasnaction link is returned and only if expand=transaction is set then the actual transaction lookup takes place

> // GET /collections/{id}
> {
>     "id": "5491f56466ad8482e8f4393be0a00052f53987346b5ae8a355bf905d35190c3e",
>     "transactions": [
>         {
>             "id": "d83a3bfe450ac7c91ce489d47678d0ea55ab29dda265a603f0b9dd6e6635234f",
>             "script": "Jlfs7ZIXBPTbAElVzMsXkoUcI6EM1pP1+PegMYHFBflKfpHqpPLS2bFApQ7wpB/bOKuRB2+I0gr4IDwMmCZjJzEj9MBWRV/ZczM6/Qk82aIJSwGHR3cYQ76ArohmK5qs+4pof84Vd+aZ4LzkQOUOzf8zvpQlt7iKt6R22qxnllgYungG3qIzNXFXhfk/pCQFt/SLcAtOIu5Eu9pbjF0Sk7RYJdHGpRLvye/pDn1rq5C6pO7PydfIgMdqJDqrto7d+6o6T/oxozJiqL27EDFhKR3i061RYZ5Sza0YW6Phur5oRXNiHJlRAkS7zg3VCnb0d3rxvqh1YezwJS6hEg6hkvsovVWfS78s/99/POTgxbIZrz/sTTLGIRy9xjKf22f/dBPJOnfkD4pnzkP/LiuuhxcOdowbE/d1k8WXCFgXUBmSpbeWRr7h5qxutMwMoZ/XsPJc2ES6wNbj2B4BpF/um7EkVDqQv+dsNZjHqqbSUIh5rq5cOR8Dz11EVdsTkisjd9umC4jUkf8BdeR5cT/j4k1JBAzkmEWf1YZS/NUF/FD8ESOB15pB6OEYfBbs8GM9ZfpyRbvdYMKDnUjmttPf2VTL9qa2Kn9QybyuL2qmPMdODF97hmlEiwIwZIlsq1clUb269kD1zITRptLZkkS37qc5NDQ=",
> 						"arguments": ["9qxmdg==", "Ej7v3+KQGenEvjUpVOe8Pg==", "0D6uxQ+ZFdg="],
> 						"reference_block_id": "2a254dd82f13557b4ea069c6ad141221374ebfda90d739fd84c9e16c86410120",
> 				    "gas_limit": 100,
> 						"payer": "3769b49731633aec",
> 						"_links": {
>                 "self": "/transactions/5491f56466ad8482e8f4393be0a00052f53987346b5ae8a355bf905d35190c3e"
>             },
>             "_expandable": {
>                 "proposal_key": "",
> 								"authorizers": "",
> 				        "payload_signatures": "",
> 				        "envelope_signatures": "",
> 				        "result": "/transactions/5491f56466ad8482e8f4393be0a00052f53987346b5ae8a355bf905d35190c3e/result"
>             }
>         }
> 				// ...
>     ],
>     "_expandable": {},
>     "_links": {
>         "self": "/collections/5491f56466ad8482e8f4393be0a00052f53987346b5ae8a355bf905d35190c3e"
>     }
> }

smnzhu:
@vishalchangrani I do understand what this tries to solve, I just don't think this is quite the right way to solve it. I would prefer one of two approaches here:

1. We keep the spec the same, but we add additional documentation to `GetCollectionByID` API description to call out that:
> It is recommended to always use `select=transaction.id` unless the transaction details are actually needed. Looking up the full transactions for a collection is expensive and will result in a slower response.

This is the preferred approach.
2. We can make the default expandable fields for a transaction different when it is returned as part of a Collection. In particular, when part of a Collection we can have only the ID field expanded by default, and all other fields should be in the `_expandable` object

vishalchangrani:
> @vishalchangrani I do understand what this tries to solve, I just don't think this is quite the right way to solve it. I would prefer one of two approaches here:
> 
> 1. We keep the spec the same, but we add additional documentation to `GetCollectionByID` API description to call out that:
> 
> > It is recommended to always use `select=transaction.id` unless the transaction details are actually needed. Looking up the full transactions for a collection is expensive and will result in a slower response.
> 
> This is the preferred approach. 2. We can make the default expandable fields for a transaction different when it is returned as part of a Collection. In particular, when part of a Collection we can have only the ID field expanded by default, and all other fields should be in the `_expandable` object

1. I actually think that people will generally not take heed of the note and will call `GetCollectionByID` without query params. The default option should be the conservative one, that avoid expensive lookup and then if the client really know what they are doing, they can pass in the `_expandable`.
2. Choosing fields of an object by `_expandable` makes it work more like the `select` field and behave differently to how other expandable fields work where the expandable refers to complete objects instead of individual fields of an object.

I view the `transactions` within a collection analogous to `payload` or `execution result` within a block response, both of which as marked as expandable. If not specified, we only return the links for payload and execution result. This is the extension of the same concept with the only difference being it is an array of expandable links instead of just one as in the case of a block response.

openapi/access.yaml
smnzhu: Without `expand=transactions`

```
{
    "id": "string",
    "_expandable": {
        "transactions": [
            "/v1/transactions/1234123124",
            "/v1/transactions/1234123125",
            "/v1/transactions/1234123126"
        ]
    },
    "_links": {
        "_self": "string"
    }
}
```

smnzhu: We can keep `_expandable` as an object rather than a list (there may be other fields than are added to `_expandable`, so I don't want to turn it into a list just for this specific API)

smnzhu: So it we decide to go forward with this change, then the convention for `_expandable` going forward will be as follows:
* If the expandable key is a sub-resource, the value should be the link to the resource
* If the expandable key is a regular field, the value should be empty string
* If the expandable key is a list of sub-resources, the value should be a list of links

smnzhu: Having said all of that, in the end I still tend to prefer approach 1: https://github.com/onflow/flow/pull/705#issuecomment-983943265

vishalchangrani: I like the this solution and have made changes for it. My reservation against approach 1 is that we are relying on optional query params to safe-gaurd against unintentional performance hit.

For point #2 - `If the expandable key is a regular field, the value should be empty string`, I was thinking of adding default values to the expandable as `""`. This helps a little in the implementation where we can stick in the default generated model for the expandable. Will create another PR for it

openapi/access.yaml
```diff
@@ -483,31 +471,25 @@ components:
...
+          type: object
+          properties:
+            transactions:
```
vishalchangrani, line 484: now _expandable has `transactions` which is an array of uris

smnzhu (approved):

PR #706, opened 2021-12-01
reverting removal require fields

vishalchangrani:
1. In my last [PR](https://github.com/onflow/flow/pull/705) I [removed](9a99bbf) the required fields from the response. However, this has a consequence of adding the `omitempty` tag for all these fields in the generated model. This results in these fields being skipped all together if not set e.g. if the `revoked` feild of the key is `false` it is not included in the response.

I could change the generated model to remove those tags but I am trying to keep the generated model as-is as much as possible.
Hence, re-introducing the required field. The onus will now be on the client to know that if a select was specified, these required fields will be skipped from the response even through they are required.

2. Changing all `uint64` integer to be of type `string` instead.

3. Finally, I noticed the [examples](https://www.notion.so/dapperlabs/REST-API-Design-9594c690c8a342e383e1d372b5860116#ff2ec5c8469748ac859b60d195ae23ac) have `"_expandable": {},`, so marking `_expandable` as require too to be always included in the response.

## Description

______

For contributor use:

### DIFF
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -440,6 +440,14 @@ components:
           $ref: '#/components/schemas/Links'
     AccountPublicKey:
       type: object
+      required:
+        - index
+        - public_key
+        - signing_algorithm
+        - hashing_algorithm
+        - sequence_number
+        - weight
+        - revoked
       properties:
         index:
           type: integer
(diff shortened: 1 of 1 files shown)


smnzhu (commented):
Not sure I understand the omitempty issue, can you explain more?

vishalchangrani:
> Not sure I understand the omitempty issue, can you explain more?

If a field is not specified as required, the code generator marks it with tag `omitempty`. This means that if a boolean field like `revoked`  is false or a the `index` feild is set to 0, it is skipped from being included in the output which is incorrect since these fields should always be included regardless of their value. Changing it to required hints the code generator (in all languages) that these fields should always be included in the response.

smnzhu (approved):

PR #713, opened 2021-12-06
Rewrite relative links so they work in github and the docs site.

10thfloor:
Relative links in GitHub markdown reference sibling pages using `./`
Because Gatsby renders files as pages using a trailing slash (default behaviour) the same link needs to be `../` resulting in broken relative links on the docs site, when importing pages from  GitHub. 

This PR adds a plugin that rewrites relative links in documents rendered from GitHub from `./` to `../` so they work in Gatsby as well.


docs/gatsby-config.js
```diff
@@ -188,7 +188,7 @@ const sources = [
     resolve: "gatsby-source-git",
     options: {
       name: "flow-js-testing-github",
-      branch: "master",
+      branch: "fix_docs",
```
kerrywei, line 191: this should be reverted

kerrywei (approved):
no concerns from my side other than the flow-js-testing branch should be restored

kerrywei (approved):

Issue #473, opened 2021-04-14
Account keys weight documentation


sideninja:
### Problem

Account [keys weight documentation](https://docs.onflow.org/concepts/accounts-and-keys/) is a bit confusing. 

It says that an account key weight is an integer between 1-1000:
> Weight (integer between 1-1000)

But then in the signing examples, it uses 0 - 1 range for weights. 
> ACCOUNT | KEY ID | WEIGHT
> -- | -- | --
> 0x01 | 1 | 1.0

Should that be changed to a number between 1 - 1000.

10thfloor:
Should be "Between 0 and 1"

bluesign:
I think should be: 

`Weight (integer between 0-1000)
`
and 

ACCOUNT | KEY ID | WEIGHT
-- | -- | --
0x01 | 1 | 1000

kerrywei:
another related PR: https://github.com/onflow/flow-go-sdk/pull/239

PR #716, opened 2021-12-08
correct key weight document to be in the range of [1, 1000]

kerrywei:
Closes: #473 

## Description

Some documents are using 0.1 and 1.0 as key weight, which should be in the range of [1, 1000]

______

For contributor use:


sideninja (approved):
Good catch

PR #718, opened 2021-12-08
add FLIP for storage API improvements in Cadence

dsainati1:
This adds a FLIP proposing a change to the account storage API in Cadence to disambiguate the cases where values are not present in storage vs. when they have an unexpected type. The implementation of this change can be found in https://github.com/onflow/cadence/pull/1286.  

______

For contributor use:

### DIFF
--- a/flips/20211208-cadence-storage-api-changes.md
+++ b/flips/20211208-cadence-storage-api-changes.md
@@ -1,0 +1,86 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
+| **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Updated**   | 2021-12-08                                           |
+
+## Objective
+
+This change would cause the `load`, `borrow`, `copy` and capability borrow 
+methods of accounts to cause a force type-cast between the type of the
+value in storage and the supplied type argument. In effect, this would
+abort execution in cases where the value the user is trying to load or
+borrow does not match the type supplied. 
+
+## Motivation
+
+The load, copy, and borrow functions, as well as the capability borrow function, 
+currently return `nil` both when the storage location is empty, and when the storage 
+location contains a value, but does not have the requested type. The inability 
+to differentiate the two cases makes debugging hard.
+
+## User Benefit
+
+This will result in less ambiguity between the two cases in which a load can fail,
+meaning users will not need to guess whether or not the value did not exist or 
+whether the type argument was wrong. 
+
+## Design Proposal
+
+This changes the interpreter's handling of storage accesses. Previously, if
+the supplied type argument did not match the value present in storage, the 
+method would return `nil`. Now, instead, it will raise a `ForceCastTypeMismatchError`
+between the expected type (the supplied type argument) and the dynamic type
+of the value. 
+
+Concretely, where this code would previously have produced `nil` for the value of `v`:
+```
+account.save(3, to: /storage/path)
+let v = account.load<String>(from: /storage/path/)
+```
+after this change, it will now abort execution.
+
+### Drawbacks
+
+This is a breaking change, and will require existing contracts to be updated to 
+account for this change. 
+
+### Best Practices
+
+This will change how users are expected to interact with storage, as they can
+no longer load values that are not guaranteed to match the supplied static type.
+Instead, users can use the `type` method to inspect the runtime type of stored values
+if they are not sure what they are, or use an `AnyStruct` or `AnyResource` type if
+they simply wish to load the value without caring what its type is. 
+
+### Compatibility
+
+This change is not backwards compatible with existing Cadence smart contracts, but
+is being made in service of a future version of Cadence wherein no backwards incompatible
+changes will be required going forwards. It should not affect any other parts of Flow.
+
+### User Impact
+
+This may break existing contracts, so there will need to be a migration path in place for this change.
+One possible option would be to release this change behind a flag that can be enabled in the emulator
+so that users can preview the effects of this change, and rewrite their contracts as necessary for
+the full release. 
+
+In general, however, there will be a general way to migrate code to be compatible with this change.
+The following code:
+```
+let v = account.load<T>(from: /storage/path)
+```
+can be rewritten as:
+```
+let v = nil
+if account.type(at: /storage/path) == Type<T>() {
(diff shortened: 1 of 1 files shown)


flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,86 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [718](https://github.com/onflow/flow/pull/718) (update when you have PR #)|
```

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,86 @@
...
+One possible option would be to release this change behind a flag that can be enabled in the emulator
+so that users can preview the effects of this change, and rewrite their contracts as necessary for
+the full release. 
+
+In general, however, there will be a general way to migrate code to be compatible with this change.
+The following code:
+```
+let v = account.load<T>(from: /storage/path)
+```
+can be rewritten as:
```
janezpodhostnik, line 77: Could an automated migration do this part somehow?

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,86 @@
...
+Concretely, where this code would previously have produced `nil` for the value of `v`:
+```
+account.save(3, to: /storage/path)
+let v = account.load<String>(from: /storage/path/)
+```
+after this change, it will now abort execution.
+
+### Drawbacks
+
+This is a breaking change, and will require existing contracts to be updated to
```
janezpodhostnik, line 48: If I understand the change correctly, the code would still work, but if an incorrect type was at the target location the script would panic instead of returning nil.

This to me does not seem to be such a big change. I would say that in general (in my un-researched opinion) developer do not rely on these methods returning nil if a different type is stored at the target location. I imagine that in general this will have no effect, but in some cases it will have the beneficial effect of detecting undesired behaviour.

janezpodhostnik (commented):
Nice! +1 for this from my side.

flips/20211208-cadence-storage-api-changes.md
bjartek: This will make it better then what it is today. The most frequent way this occurs IMHO is when people use the same contract under different addresses. 

The error message will tell what contract the type is from if it is from userland?

flips/20211208-cadence-storage-api-changes.md
dsainati1: In theory, but I think it's likely this will not be necessary in most cases, and this would double the number of storage queries if we did this for every case. It's only required in places where the stored type might not be the same as the loaded type, so I think it would be better to be manual in only those cases where it is required.

flips/20211208-cadence-storage-api-changes.md
dsainati1: I agree that it is unlikely to significantly affect any contracts, but developers will have to audit their contracts to make sure they don't crash as a result of the change.

janezpodhostnik: >  but developers will have to audit their contracts to make sure they don't crash as a result of the change

Agreed! I was jus wondering if it is worth reiterating in this chapter, that while it is a breaking change, and you should re-check your contract, it is likely that no action (other than re-checking) will be needed.

joshuahannan (approved):
So it will still always return an optional and return`nil` when there is nothing there? Thats what it looks like to me. If that is the case, this looks good to me! 😃

janezpodhostnik (approved):

bluesign:
Seems very good proposal 👍

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,89 @@
+# Title of FLIP
```
dsainati1, line 1:
```suggestion
# Cadence Storage API Improvements
```

SupunS (approved):
I mostly reviewed it from the cadence's side of things and looks good! 👌

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,89 @@
+# Cadence Storage API Improvements
+
+| Status        | (Proposed)       |
```
dsainati1, line 3:
```suggestion
| Status        | (Approved)       |
```

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,89 @@
...
+This changes the interpreter's handling of storage accesses. Previously, if
+the supplied type argument did not match the value present in storage, the 
+method would return `nil`. Now, instead, it will raise a `ForceCastTypeMismatchError`
+between the expected type (the supplied type argument) and the dynamic type
+of the value. 
+
+Concretely, where this code would previously have produced `nil` for the value of `v`:
+```
+account.save(3, to: /storage/path)
+let v = account.load<String>(from: /storage/path/)
```
turbolent, line 42:
```suggestion
account.save(3, to: /storage/path)
let v = account.load<String>(from: /storage/path)
```

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,89 @@
...
+### User Impact
+
+This may break existing contracts, so there will need to be a migration path in place for this change.
+One possible option would be to release this change behind a flag that can be enabled in the emulator
+so that users can preview the effects of this change, and rewrite their contracts as necessary for
+the full release. 
+
+In general, however, there will be a general way to migrate code to be compatible with this change.
+The following code:
+```
```
turbolent, line 77: Same here, maybe add a newline before/after and add the `cadence` language tag

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,89 @@
...
+let v = account.load<T>(from: /storage/path)
+```
+can be rewritten as:
+```
+let v = nil
```
turbolent, line 82: This variable will get the inferred `Never?`, I think this needs a type annotation.
Also, the variable is `let` and is assigned to in the if statement, so it should be `var`.

```suggestion
var v: T? = nil
```

flips/20211208-cadence-storage-api-changes.md
```diff
@@ -0,0 +1,89 @@
...
+if account.type(at: /storage/path) == Type<T>() {
+    v = account.load<T>(from: /storage/path)
+}
+```
```
turbolent, line 86: Maybe this should be provided in a function? It might reduce the repetition of the static type (2x) and the path (2x), and the intent would be easier to reason about  I can't quite think of a good name though

PR #722, opened 2021-12-09
Add FLIP for making references to indexed accesses optional values

dsainati1:
This adds a FLIP outlining a proposal to change the functionality of creating references to the results of indexed accesses in Cadence to produce an optional reference. The implementation of this FLIP is in https://github.com/onflow/cadence/pull/1303. 

Fixes: https://github.com/onflow/cadence/issues/747

______

For contributor use:


flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,114 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [722](https://github.com/onflow/flow/pull/722) (update when you have PR #)|
```

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,88 @@
+# Title of FLIP
```
dsainati1, line 1:
```suggestion
# Optional References to Indexed Accesses
```

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,88 @@
...
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [722](https://github.com/onflow/flow/pull/722) (update when you have PR #)|
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [722](https://github.com/onflow/flow/pull/722) |
```

joshuahannan (approved):
Looks good to me!

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,88 @@
...
+
+## Motivation
+
+As detailed in https://github.com/onflow/cadence/issues/747, it is possible to 
+crash execution with a failed dereference that is not caught by the typechecker
+by creating a reference to the result of an indexed access, and then accessing
+a field on that reference. This occurs because while references to optional 
+values are normally disallowed, references to indexed accesses are special cased
+in the checker. In order to proceed with a stable release of Cadence, this bug 
+needs to be addressed.
```
turbolent, line 25: This was implemented specifically as a feature and is not a bug

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,88 @@
...
+}
+let dict: {String : S} = {}
+let s = &dict[""] as &S // s has type &S?
+let n: Number = s.foo // optional does not have member "foo"
+```
+
+### Drawbacks
+
+This is a breaking change, and will require existing contracts to be updated to 
+account for this change.
```
turbolent, line 69: How can existing code be updated? For example, many NFT collections rely on this functionality, see e.g.
- https://github.com/dapperlabs/nba-smart-contracts/blob/ceeb3d5d9c6238bbbd9e153171368c1031275e59/contracts/TopShot.cdc#L558
- https://github.com/dapperlabs/nba-smart-contracts/blob/ceeb3d5d9c6238bbbd9e153171368c1031275e59/contracts/TopShot.cdc#L711

Maybe add an example

turbolent (commented):
Currently the proposal only proposes the removal of references to optional values.
Should it also propose the addition of support for optional references?
Specifically, https://github.com/onflow/cadence/pull/1303 implements a certain new feature where `foo[bar] as &T` results in`&T?` (interpreted as `(&T)?`). 

I think it makes sense to add support for optional references as a replacement for referenes to optional values, but unlike in the PR, I'd suggest that:
- an index expression which returns an optional, e.g. for a dictionary, should require the target optional type to be optional, e.g. 
    ```rust
    let xs: {Int: Int} = {0: 1}
    let ref = &xs[0] as &Int?
    ```
- an index expression which does not return an optional, e.g. for arrays, should not result in an optional reference (I think the current implementation of the PR does)
  ```rust
  let xs = [1]
  let ref = &xs[0] as &Int
  ```

dsainati1:
> Should it also propose the addition of support for optional references?

As far as I am aware, these already existed in Cadence. It is currently possible to get an optional reference value by looking up a value in a dictionary of references, for example:

```cadence
let x = &3 as &Int
let y = {"":x}
let z = y[""]
```

or by returning a reference or `nil` from a function:

```cadence
pub fun foo(): &Int? {
    if cond {
       return &3 as &Int
    } 
    return nil
}
```

I think this FLIP and its associated PR do not introduce anything that doesn't already exist into the language, only that we replace the old references to optional with optional references. 

That said, I prefer your description of the implementation to what is currently present in the PR, so I will update the PR to reflect this better design.

dete (approved):
Makes sense to me!

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,88 @@
+# Optional References to Indexed Accesses
+
+| Status        | (Proposed)       |
```
dsainati1, line 3:
```suggestion
| Status        | (Approved)       |
```

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,116 @@
...
+only way to obtain a reference to an optional, and as such this change effectively 
+removes these references from the language entirely. Instead, it allows users to 
+make the type on the right-hand side of an `as` expression an optional, and 
+allows users to create optionally-typed references in place of references
+to optional values. 
+
+## Motivation
+
+As detailed in https://github.com/onflow/cadence/issues/747, it is possible to 
+crash execution with a failed dereference that is not caught by the typechecker
```
turbolent, line 23: Maybe don't use the word "crash" here: Program execution aborts / stops, but the execution environment does not crash

```suggestion
abort execution with a failed dereference that is not caught by the type checker
```

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,116 @@
...
+by creating a reference to the result of an indexed access, and then accessing
+a field on that reference. This occurs because while references to optional 
+values are normally disallowed, references to indexed accesses are special cased
+in the checker. In order to proceed with a stable release of Cadence, this feature
+needs to be removed. 
+
+## User Benefit
+
+This removes an unsound behavior from the type checker, preventing an error
+from occurring at runtime.
```
turbolent, line 33: The current feature is not unsound behaviour, it was intentionally added. At the time it was added there were two options, references to optionals and optional references, and we opted for the current former. The proposal suggests the other option, optional references.

The user benefit is still true: The proposed behaviour removes the potential for the program to unexpectedly abort due to the implicit unwrap of the references optional failing.

flips/20211208-cadence-remove-optional-reference.md
```diff
@@ -0,0 +1,116 @@
...
+
+```cadence
+pub struct S {
+    pub let foo : Number
+    init() {
+        self.foo = 0
+    }
+}
+let dict: {String: S} = {}
+let s = &dict[""] as &S? // s has type &S?
```
turbolent, line 66: Maybe document here that `&S?` reads `(&S)?`, i.e. the result of the reference expression is going to be an optional, and not `&(S?)` (the result is a reference to an optional)

flips/20211208-cadence-remove-optional-reference.md
dsainati1: Is this not still unsound behavior, even if it is intentional? This makes it possible to obtain a value that has a runtime type that differs from its static type, which feels like a pretty straightforward example of unsoundness.

turbolent:
Regarding the current implementation: I just checked, and `&T?` is already parsed as `(&T)?` today 👍

turbolent (approved):

PR #727, opened 2021-12-11
Updates FLIP 200 with new proposal updates

JeffreyDoyle:
## Description

Updates FLIP 200 with new proposal updates

### DIFF
--- a/flips/20201125-bip-44-multi-account.md
+++ b/flips/20201125-bip-44-multi-account.md
@@ -153,8 +153,8 @@ The prescribed account discovery procedure is as follows:
    Note: If you are unable to derive keys for a certain elliptic curve, 
    you may fail to discover an account / key that the user has generated 
    previously.
-1. Scan for each keys usage by querying the public key registry with the key,
-   curve and hash algorithm for it.
+1. Scan for each keys usage by querying the public key registry with the key and
+   elliptic curve used to generate it.
    If the gap limit has been reached for both key index and account index, stop
    discovery.
     - 3.1. If no addresses are found in the registry
--- a/flips/20201125-bip-44-multi-account.md
+++ b/flips/20201125-bip-44-multi-account.md
@@ -143,8 +143,8 @@ The prescribed account discovery procedure is as follows:
    query the Flow network to fetch the account's details. If an account 
    is found, remember the relationship between the path used to generate 
    this key and the account's details.
-2. Derive the key pair(s) (starting with account index = 0 and key index =
-   1) using the path specified in this FLIP and each elliptic curve that
+2. Derive the key pair(s) (starting with account index = 0 and key index = 1) 
+   using the path specified in this FLIP and each elliptic curve that
    both you and Flow support (Flow currently supports keys generated on the
    ECDSA P256<sup>1</sup> and ECDSA secp256k1<sup>2</sup> curves) for each 
    hash algorithm that both you and Flow support (Flow currently supports 
--- a/flips/20201125-bip-44-multi-account.md
+++ b/flips/20201125-bip-44-multi-account.md
@@ -122,6 +122,9 @@ The legacy path is:
 m / 44' / 1' / 769 / 0 / 0
 ```
 
+All keys generated with the legacy path as of today (Dec 2021) have been done so using the
+ECDSA P-256 curve.
+
 ### Account Discovery
 
 Wallet developers should use a modified version of the original account
@@ -135,29 +138,52 @@ The account index gap limit should be `5`.
 
 The prescribed account discovery procedure is as follows:
 
-1. Derive the key pair using the legacy path, checking it's use with the
-   public key registry. If an address is found, query the Flow network to fetch
-   the account's details. If an account is found, remember the relationship between the
-   path used to generate this key, and the account's details.
-2. Derive a key pair (starting with account index = 0 and key index =
-   1) using the path specified in this FLIP.
-3. Scan for the key usage by checking the public key registry.
+1. Derive the key pair using the legacy path and ECDSA P256<sup>1</sup>, 
+   checking its use with the public key registry. If an address is found, 
+   query the Flow network to fetch the account's details. If an account 
+   is found, remember the relationship between the path used to generate 
+   this key and the account's details.
+2. Derive the key pair(s) (starting with account index = 0 and key index =
+   1) using the path specified in this FLIP and each elliptic curve that
+   both you and Flow support (Flow currently supports keys generated on the
+   ECDSA P256<sup>1</sup> and ECDSA secp256k1<sup>2</sup> curves) for each 
+   hash algorithm that both you and Flow support (Flow currently supports 
+   SHA2_256<sup>3</sup>, SHA2_384<sup>4</sup>, SHA3_256<sup>5</sup>, SHA3_384
+   <sup>6</sup> hashing algorithms).
+3. Scan for each keys usage by querying the public key registry with the key,
+   curve and hash algorithm for it.
    If the gap limit has been reached for both key index and account index, stop
    discovery.
-    3.1. If no address is found in the registry 
-        3.1.1 If the key index gap limit has been reached without finding any 
+    - 3.1. If no addresses are found in the registry
+        - 3.1.1 If the key index gap limit has been reached without finding any
         addressed in the registry, then go to step 2, incrementing the account index
-        by one and starting with key index = 0 again. 
-        3.1.2 If the key index gap limit has not been reached, go to step 2 and
-        increment the key index by one. 
-    3.2. If an address is found, query the Flow network to fetch the account's details. 
-        3.2.1. If no account is found<sup>1</sup>, go to step 2, incrementing 
+        by one and starting with key index = 0 again.
+        - 3.1.2 If the key index gap limit has not been reached, go to step 2 and
+        increment the key index by one.
+    - 3.2. If one or more addresses are found, query the Flow network to fetch each
+      account's details.
+        - 3.2.1. If no accounts are found<sup>7</sup>, go to step 2, incrementing
         the account index by one.
-        3.2.2. If an account is found, remember the relationship between the
-        path used to generate this key and the account's details. Then go to step 2, 
-        incrementing the key index by one.
+        - 3.2.2. For each account found, remember the relationship between the path
+        used to generate this key, the curve used to generate this key,
+        and the hash algorithm corresponding to this key and the account's details.
+        Then go to step 2, incrementing the key index by one.
 
-<sup>1</sup>Flow supports account deletion, meaning that an address found in the
+<sup>1</sup>ECDSA P256 is Elliptic Curve Digital Signature Algorithm (ECDSA) on
+the NIST P-256 curve.
+
+<sup>2</sup>ECDSA secp256k1 is Elliptic Curve Digital Signature Algorithm (ECDSA) on
+the secp256k1 curve.
+
+<sup>3</sup>SHA2_256 is Secure Hashing Algorithm 2 (SHA-2) with a 256-bit digest.
+
+<sup>4</sup>SHA2_384 is Secure Hashing Algorithm 2 (SHA-2) with a 384-bit digest.
+
+<sup>5</sup>SHA3_256 is Secure Hashing Algorithm 3 (SHA-3) with a 256-bit digest.
+
+<sup>6</sup>SHA3_384 is Secure Hashing Algorithm 3 (SHA-3) with a 384-bit digest.
+
+<sup>7</sup>Flow supports account deletion, meaning that an address found in the
 registry may refer to a nonexistent account. In this case the account should be
 skipped but discovery should continue.
 
@@ -193,6 +219,20 @@ generate the additional key should be:
 When generating additional keys for an account, the correct key index to
 use in the path to generate the new key should be the smallest unused key index
 available which has not yet been used to generate a key on the account.
+
+##### Conflict Resolution: Multiple Signature & Hashing Algorithms Used for Keys Generated Using the Same Path
+
+The same derivation path should not be used with multiple combinations of signature and
+hashing algorithms to generate keys set on an account.
+
+The prescribed account discovery algorithm will find keys generated using
+all combinations of derivation paths, supported signature algorithms and supported hashing
+algorithms. However, wallets should not use the same path twice to apply
+it to multiple signature and hashing algorithms to generate keys.
+
+When generating a new key pair for an account, a new path, which has not yet been used in
+combination with any signature or hashing algorithm to generate keys set on the account,
+should be used.
 
 ### Drawbacks
 
--- a/flips/20201125-bip-44-multi-account.md
+++ b/flips/20201125-bip-44-multi-account.md
@@ -67,23 +67,18 @@ two account indices should be used to generate child public keys that are
 assigned to the same Flow account address. Furthermore, a single account index
 should not be used to generate keys for more than one account address.
 
-The Account Index to use when generating a key pair to be added on a new account
-is the smallest Account Index of which has not yet been used to generate keys 
-set on any account.
-
 The motivation for these accounts is the same set forth in BIP 44:
 
 > Users can use these accounts to organize the funds in the same fashion as bank
 > accounts; for donation purposes (where all addresses are considered public),
 > for saving purposes, for common expenses etc.
 
-Unlike BIP 44, wallets should not prevent the creation of a new account if any
-existing account corresponding to a previous Account Index does not have any
-transaction history.
+When generating a key pair to be added on a new account, use the smallest 
+account index that has not already been used to generate keys for an existing account.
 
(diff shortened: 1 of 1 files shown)


flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
-to the same Flow account address. Furthermore, a single account index should not
-be used to generate keys for more than one account address.
+Account is the index of a distinct account address owned by the wallet user. No
+two account indices should be used to generate child public keys that are
+assigned to the same Flow account address. Furthermore, a single account index
+should not be used to generate keys for more than one account address.
+
+The Account Index to use when generating a key pair to be added on a new account
+is the smallest Account Index of which has not yet been used to generate keys 
+set on any account.
```
psiemens, line 72:
```suggestion
When generating a key pair for a new account, use the smallest Account Index that has not already been used to generate keys for an existing account.
```

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
-Account is the index of a distinct account address owned by the wallet user. 
-No two account indices should be used to generate child public keys that are assigned 
-to the same Flow account address. Furthermore, a single account index should not
-be used to generate keys for more than one account address.
+Account is the index of a distinct account address owned by the wallet user. No
+two account indices should be used to generate child public keys that are
+assigned to the same Flow account address. Furthermore, a single account index
+should not be used to generate keys for more than one account address.
+
+The Account Index to use when generating a key pair to be added on a new account
```
psiemens, line 70:
```suggestion
The account index to use when generating a key pair to be added on a new account
```

Personally I'd like to avoid capitalizing terms that aren't proper nouns. I find it difficult to read and don't think it adds much clarity.

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
 
-### Account Discovery
+When generating a new key pair for an account, the Key Index to use in the path to 
+generate the new key should be the smallest unused Key Index available which 
+has not yet been used to generate a key on the account.
+
+### Legacy Path
+
+There exist Flow accounts with keys on them which have been generated using the
+Legacy Path.
```
psiemens, line 115: I think you can be explicit here and mention that, up until now, the Ledger application created Flow accounts using a different interpretation of BIP 44 that we now refer to as the legacy path. And to our knowledge, this is the only application that created accounts in this way.

Great idea to mention the legacy path here!

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
+        the Account Index by one.
+        3.2.2. If an account is found, remember the relationship between the
+        path used to generate this key and the account's details. Then go to step 2, 
+        incrementing the Key Index by one.
+
+<sup>1</sup>Flow supports account deletion, meaning that an address found in the
+registry may refer to a nonexistent account. In this case the account should be
+skipped but discovery should continue.
+
+#### Account Discovery Conflict Resolution
```
psiemens, line 165: Nice, I like this solution.

psiemens (commented):
Nice @JeffreyDoyle!

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
-No two account indices should be used to generate child public keys that are assigned 
-to the same Flow account address. Furthermore, a single account index should not
-be used to generate keys for more than one account address.
+Account is the index of a distinct account address owned by the wallet user. No
+two account indices should be used to generate child public keys that are
+assigned to the same Flow account address. Furthermore, a single account index
+should not be used to generate keys for more than one account address.
+
+The Account Index to use when generating a key pair to be added on a new account
+is the smallest Account Index of which has not yet been used to generate keys
```
Wolog2021, line 71: Suggest changing "Account Index of which" to "Account Index which" i.e. remove "of"

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
+set on any account.
 
 The motivation for these accounts is the same set forth in BIP 44:
 
-> Users can use these accounts to organize the funds in the same fashion as bank accounts; for donation purposes (where all addresses are considered public), for saving purposes, for common expenses etc.
+> Users can use these accounts to organize the funds in the same fashion as bank
+> accounts; for donation purposes (where all addresses are considered public),
+> for saving purposes, for common expenses etc.
+
+Unlike BIP 44, wallets should not prevent the creation of a new account if any
```
Wolog2021, line 80: I might rephrase this for clarity as follows:

Key difference: BIP 44 requires that wallets check the accounts associated with previous account indices. If a previous account does not have a transaction history, creation of a new account is prevented. This is expressly not a requirement in this specification (FLIP-200).

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
+any key.
 
-### Account Discovery
+When generating a new key pair for an account, the Key Index to use in the path to 
+generate the new key should be the smallest unused Key Index available which 
+has not yet been used to generate a key on the account.
+
+### Legacy Path
+
+There exist Flow accounts with keys on them which have been generated using the
```
Wolog2021, line 114: Is there a more formal term we could consider than "on them" such as "associated" or "assigned" that would make sense?

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
+The Key Index gap limit should be `5`.
+The Account Index gap limit should be `5`.
+
+The prescribed Account Discovery procedure is as follows:
+
+1. Derive the key pair using the Legacy Path, checking it's use with the
+   public key registry. If an address is found, query the Flow network to fetch
+   the account's details. If an account is found, remember the relationship between the
+   path used to generate this key, and the account's details.
+1. Derive a key pair (starting with Account Index = 0 and Key Index =
```
Wolog2021, line 143: Step numbering through here appears to be off by one.

flips/20201125-bip-44-multi-account.md
```diff
@@ -9,188 +9,279 @@
...
+
+If a wallet has used the Account Index field incorrectly to generate keys for a
+user, conflict resolution logic may be required when attempting to generate new
+keys.
+
+When generating additional keys, the correct Account Index to use in the path to
+generate the additional key should be:
+
+- If adding a key to a new account:
+  - The smallest Account Index of which has not yet been used to generate keys
```
Wolog2021, line 183: Suggest removing the word "of"

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+
+- If adding a key to a new account:
+  - The smallest account index which has not yet been used to generate keys
+    set on any account.
+- If adding a key to an existing account:
+  - The largest account index that account discovery determines has been
+    previously used to generate existing keys set on the account, which has also
+    not been used incorrectly to generate keys set on another account.
+  - Otherwise, the smallest account index which has not yet been used to
+    generate keys set on any account.
```
tarakby, line 189: I think I see what this sentence means 👌🏼  but I'm sure examples will make it clearer 😅   I'm not saying the examples need to be added to this document.

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+
+##### Conflict Resolution: Incorrect Use of Key Index
+
+When generating additional keys for an account, the correct key index to
+use in the path to generate the new key should be the smallest unused key index
+available which has not yet been used to generate a key on the account.
```
tarakby, line 195: This is great, and should be enough. 
To avoid confusion we could also add that a path used with at least one curve is considered "used".  This would prevent wallets saying "the path has been used with curve_1, but it hasn't been used with key_2 yet, so I'm gonna use it again".

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+    3.2. If an address is found, query the Flow network to fetch the account's details. 
+        3.2.1. If no account is found<sup>1</sup>, go to step 2, incrementing 
+        the account index by one.
+        3.2.2. If an account is found, remember the relationship between the
+        path used to generate this key and the account's details. Then go to step 2, 
+        incrementing the key index by one.
+
+<sup>1</sup>Flow supports account deletion, meaning that an address found in the
+registry may refer to a nonexistent account. In this case the account should be
+skipped but discovery should continue.
```
tarakby, line 162: This is great! Should we also add a similar note for revoked keys? I mean we could precise if we consider a path leading to a revoked key, as a "used" path or a "free" path.

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
 ## Design Proposal
 
 ### Path Levels
 
-BIP 44 defines an implementation of 
-[Bitcoin Improvement Proposal 32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) (BIP 32)
-that standardizes a format for hierarchical deterministic (HD) 
-key derivation paths that support multiple coins and multiple accounts per coin.
+BIP 44 defines an implementation of [Bitcoin Improvement Proposal
+32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki) (BIP 32)
```
tarakby, line 39: The issue with BIP44 and BIP32 is that they only define the derivation on curve secp256k1. 
Referencing BIP32 shouldn't be enough to apply the derivation to another curve, unless we define how we want to generalize the BIP32 technique to other curves. 

There is a [SLIP](https://github.com/satoshilabs/slips/blob/master/slip-0010.md) that did something similar  "SLIP-0010 describes how to derive private and public key pairs for curve types different from secp256k1." but I'm having minor concerns with it. 
 I'll reach out to Vacuum Labs and maybe Ledger to see if there is a known generalization we can reference, otherwise we might need to describe the generalization ourselves (not long). I'll follow up on this.

tarakby (commented):
Nicely written update 👏🏼

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
-2. Scan for addresses by checking each public key index against the public key registry,
-respecting the [gap limit described in BIP 44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#address-gap-limit).
-3. If no address is found in the registry, stop discovery.
-4. If an address is found, query the chain to fetch the account details.
-5. If no account is found<sup>1</sup>, skip this index and go to step 1.
-6. If an account is found, add it to the user's wallet and go to step 1.
+The legacy path is:
+
+```
+m / 44' / 1' / 769 / 0 / 0
```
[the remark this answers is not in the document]
psiemens: I agree with Tarak's suggestion. All keys created to date have been on the P-256 curve, so let's specify that alongside the legacy path.

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+44](https://github.com/bitcoin/bips/blob/master/bip-0044.mediawiki#account-discovery).
+
+Modified constants from BIP 44 include:
+
+The key index gap limit should be `5`.
+The account index gap limit should be `5`.
+
+The prescribed account discovery procedure is as follows:
+
+1. Derive the key pair using the legacy path, checking it's use with the
```
psiemens, line 138:
```suggestion
1. Derive the key pair using the legacy path, checking its use with the
```

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+   public key registry. If an address is found, query the Flow network to fetch
+   the account's details. If an account is found, remember the relationship between the
+   path used to generate this key, and the account's details.
```
psiemens, line 141:
```suggestion
   path used to generate this key and the account's details.
```

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+2. Derive a key pair (starting with account index = 0 and key index =
+   1) using the path specified in this FLIP.
+3. Scan for the key usage by checking the public key registry.
+   If the gap limit has been reached for both key index and account index, stop
+   discovery.
+    3.1. If no address is found in the registry 
+        3.1.1 If the key index gap limit has been reached without finding any
```
psiemens, line 148: I just noticed that the formatting doesn't work for these sub-steps when viewing the rendered markdown (it all goes on one line). Maybe try adding a tick (`-`) before each?

flips/20201125-bip-44-multi-account.md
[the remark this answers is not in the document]
JeffreyDoyle: @tarakby Do we need to be concerned with BLS keys? Or just `ECDSA_P256` and `ECDSA_secp256k1` ?

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+   path used to generate this key, and the account's details.
+2. Derive a key pair (starting with account index = 0 and key index =
+   1) using the path specified in this FLIP.
+3. Scan for the key usage by checking the public key registry.
+   If the gap limit has been reached for both key index and account index, stop
+   discovery.
+    3.1. If no address is found in the registry 
+        3.1.1 If the key index gap limit has been reached without finding any 
+        addressed in the registry, then go to step 2, incrementing the account index
+        by one and starting with key index = 0 again.
```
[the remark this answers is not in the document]
JeffreyDoyle: We introduce the account index gap limit before the discovery algorithm. I want to keep the termination step for the algorithm in step 3, because the terminal state for the algorithm is when both the account index gap limit, and the key index gap limit have been reached

flips/20201125-bip-44-multi-account.md
JeffreyDoyle: Added a new conflict resolution section for this purpose 👍

flips/20201125-bip-44-multi-account.md
```diff
@@ -5,192 +5,282 @@
...
+The key index gap limit should be `5`.
+The account index gap limit should be `5`.
+
+The prescribed account discovery procedure is as follows:
+
+1. Derive the key pair using the legacy path, checking it's use with the
+   public key registry. If an address is found, query the Flow network to fetch
+   the account's details. If an account is found, remember the relationship between the
+   path used to generate this key, and the account's details.
+2. Derive a key pair (starting with account index = 0 and key index =
```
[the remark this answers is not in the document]
JeffreyDoyle: Updated to include iterating over both the supported curves, and hashing algorithms as well

flips/20201125-bip-44-multi-account.md
JeffreyDoyle: I think it might depend on how the Public Key registry / Hardware Wallet API will have to work. If it returns back addresses for keys that are on the account but are also revoked, i'm less worried about confusion here than if it doesn't return addresses for such keys.

flips/20201125-bip-44-multi-account.md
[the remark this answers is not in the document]
tarakby: No, BLS is not supported by Flow accounts (BLS is just supported by Cadence, that's why you might see it in Cadence docs).

flips/20201125-bip-44-multi-account.md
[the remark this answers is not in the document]
tarakby: You're right, the account gap is already defined earlier 👌🏼  
What I meant is that the algorithm (steps 1. 2. 3.) does not describe what to do with the account gap (while it describes how to deal with the key gap).

flips/20201125-bip-44-multi-account.md
tarakby: Yep good point, I think this could be precised on the key registry side.

flips/20201125-bip-44-multi-account.md
tarakby: The new section is great 👌🏼

flips/20201125-bip-44-multi-account.md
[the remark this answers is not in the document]
JeffreyDoyle: Ah yes, the account gap limit is increased when the algorithm hits step 3.1.1

psiemens (approved):
Looks good to me!

flips/20201125-bip-44-multi-account.md
tarakby: Update: I reached out to Ledger, but I looks like the answer will come after the holidays. Their answer should help us update FLIP200 (the current doc cannot be complete without specifying this curve derivation point unfortunately) 

cc @JeffreyDoyle

psiemens: Thanks @tarakby. Let's move this discussion to #200 so it doesn't get lost

PR #728, opened 2021-12-13
Add feature parity with gRPC

sideninja:
Closes: [flow-go/1762](https://github.com/onflow/flow-go/issues/1762)

## Description
Adding endpoints to match features of the gRPC. Added endpoints are:
- Event endpoint handling event for block height range or block IDs with provided type filter
- Status endpoint which covers two requests in gRPC API - ping and network

Small fix: change status to string as defined in spec.
______

For contributor use:


smnzhu (commented):
Can you add more detail about why the status is changed to string?

openapi/access.yaml
```diff
@@ -373,10 +373,82 @@ paths:
...
+          description: OK
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Status'
+        '404':
+          $ref: '#/components/responses/404NotFound'
+        '500':
+          $ref: '#/components/responses/500InternalServerError'
+      operationId: GetNetworkStatus
```
smnzhu, line 396: I would highly recommend we drop this endpoint, because block headers already contain the Chain ID so this is very redundant: https://github.com/onflow/flow-go/blob/c5a986b8588b1891c059e77656a8f831d90742aa/model/flow/header.go#L20

openapi/access.yaml
```diff
@@ -373,10 +373,82 @@ paths:
...
+          explode: false
+          allowEmptyValue: false
+          schema:
+            type: array
+            items:
+              $ref: '#/components/schemas/Identifier'
+            minItems: 1
+            maxItems: 50
+            uniqueItems: true
+        - $ref: '#/components/parameters/selectParam'
```
smnzhu, line 451: I would also recommend against having this endpoint, this was one that I purposely omitted when I was designing the API. There are a couple reasons:

* First, it is already possible to achieve a query for the same underlying data by querying the `/blocks` API by height range or ids, and then `expand`ing / `select`ing `execution_result.events`. If you look at the gRPC implementation for the `GetEventsBy...`  API's, this is exactly what happens underneath the hood.
* This API is not RESTful in nature, because events are not actually uniquely identifiable resources, and they only make sense within the context of a block or transaction. 

When we talk about feature parity, what we really mean is that we want to be able to do all the things we could do with the previous API using this API. This doesn't mean that we need to have all the same endpoints, or that things need to be queried the same way. Otherwise, it would have been very easy to simply design a JSON version of the gRPC API, or some other JSON RPC api. 

However, we specifically decided to design a REST API, and the existing gRPC API doesn't have RESTful semantics, so naturally there will be some differences in how they should be queried.

smnzhu (changes requested):

openapi/access.yaml
sideninja: You are right, I agree.

openapi/access.yaml
sideninja: I understand that feature parity is not having the same endpoints but the ability to achieve the same functionality. Although honestly, I overlooked this possibility of having events as part of blocks I still see a potential problem with the suggested approach for fetching events, that is, inability to query by the event type on the backend, I'm afraid that if our suggested option is for devs to fetch all events in big ranges of a block to only then filter them on the client-side it will generate a lot of data being sent over the network, don't you think? I imagine a lot of transactions in a block can be token transfers that have a lot of events and then fetching that for ranges of 50 blocks will amount to a lot of data. This is not necessarily an argument for my approach but more of a concern with the current suggestion. Adding this filter to block also doesn't really feel RESTy. What is your take on this?

openapi/access.yaml
vishalchangrani: I have removed the `/status` endpoint

openapi/access.yaml
vishalchangrani: retaining this as per our [slack discussion](https://axiomzen.slack.com/archives/C02F8LFCCP8/p1639590253006600)

vishalchangrani (approved):

smnzhu (approved):

PR #729, opened 2021-12-13
Add FLIP for improving Cadence numeric supertypes

SupunS:
Work towards https://github.com/dapperlabs/cadence-private-issues/issues/10

## Description

Add a FLIP for improving arithmetic operations on numeric supertypes in cadence.

______

For contributor use:


flips/20211210-cadence-numeric-supertypes.md
```diff
@@ -0,0 +1,118 @@
...
+- SignedFixedPoint
+- Integer
+- SignedInteger
+
+Currently, the arithmetic operations are also allowed on those types (e.g. `+`, `-` `*`, `/`, etc.).
+
+However, it is not always guaranteed that these operations will succeed at run-time.
+
+e.g:
+If the two numeric values are of same type, then the operation will succeed.
```
turbolent, line 32:
```suggestion
For example, if the two numeric values are of same type, then the operation will succeed:
```

flips/20211210-cadence-numeric-supertypes.md
```diff
@@ -0,0 +1,118 @@
...
+```cadence
+let x: Integer = 3 as Int8
+let y: Integer = 4 as Int8
+let z: Integer = x + y
+```
+If the two numeric values are of different types, then there will be a run-time panic.
+```cadence
```
turbolent, line 39:
```suggestion

However, if the two numeric values are of different types, then there will be a run-time panic.
```cadence
```

flips/20211210-cadence-numeric-supertypes.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Design Proposal
+
+The proposed solution is to statically disallow arithmetic operations on numeric supertypes.
+```
+let x: Integer = 3 as Int8
+let y: Integer = 4 as Int8
+
+let z: Integer = x + y     // Semantic error
+```
```
turbolent, line 64:
```suggestion

```cadence
let x: Integer = 3 as Int8
let y: Integer = 4 as Int8

let z: Integer = x + y     // Static error
```
```

File: flips/20211210-cadence-numeric-supertypes.md [hunk 4]
```diff
@@ -0,0 +1,118 @@
...
+
+Developers will have to explicitly force-cast it to the desired type before they do the arithmetic
+operation.
+```
```
turbolent, line 68:
```suggestion

```cadence
```

turbolent (approved):
+1 from me. 

I think removing the operations from the supertypes and having static guarantees is the best option.
The alternative to define run-time behaviour for arithmetic operations of different types seems unintuitive and error-prone.

dsainati1:
+1 from me as well.

I think there is potentially room for a middle ground between disallowing the behavior on all supertypes and implementing different behavior at runtime for every combination of runtime types. One option might be to allow `Integer`s to be added, subtracted or multiplied (but not divided) with each other (using the size of the larger of the two integer types at runtime as the result type), since we know that adding, subtracting and multiplying integers will always produce an integer. 

However it's not clear to me that this complexity is necessarily worth it. This approach will at least put us in a situation where we can add new behavior in later if we want to, since banning the behavior now ensures that adding new behavior in later will not be a breaking change.

flips/20211210-cadence-numeric-supertypes.md
joshuahannan: Shouldn't one of those be a different type? Why would two `Int8` numbers being added cause a static error?

flips/20211210-cadence-numeric-supertypes.md
```diff
@@ -0,0 +1,118 @@
...
+let x: Integer = 3 as Int8
+let y: Integer = 4 as Int8
```
joshuahannan, line 70: Should one of these be a different type?

joshuahannan (commented):
I don't know if I'm fully on board with this yet. I almost feel like more descriptive error messages could solve this problem better than restricting the behavior of the supertypes. This definitely makes the supertypes less useful. To be fair though, I haven't ever used this feature of supertypes before so it clearly isn't that useful in the first place.

flips/20211210-cadence-numeric-supertypes.md
SupunS: This demonstrates the new behavior once the change is made. This example is for the successful scenario, where runtime values are of the same type.

To clarify more: Here even though the two values `3` and `4` are of `Int8` type, the two variables `x` and `y` are of `Integer` supertype. Hence the `x + y` is not allowed. Instead, to make it work, users will have to cast it before add-operation (`(x as! Int8) + (y as! Int8)`). Then it would work as before.

SupunS:
@joshuahannan Currently, when two number supertypes are added, say `x + y`, where both `x` and `y` are `Integer` type, then there is a force-cast that happens under-the-hood at runtime, to make sure runtime types of the two values are same. 
The argument against this current behavior was that it's not intuitive enough that the code could cause runtime panics. Its sort of hidden from the developers/readers of the code.

This proposal is to: rather than runtime doing that cast under the hood, make users do that cast explicitly. So they would know that there is an unsafe operation in the code. It aligns more with the other language semantics, and avoids unexpected panics on arithmetic operations.

And yes, as you've also mentioned, arithmetic operations on supertypes might rarely be used in the real-world, so the impact would also be quite minimum.

SupunS:
I updated the proposal to include suggestions for the comparison operations as well. Please have another look if you already reviewed the first version. Thank you! 🙏

dete:
>  Numeric supertypes may still be useful to have for certain use-cases. e.g: collections, etc.

Can you clarify this? It seems to me that this proposal makes the numeric supertypes really quite unnnecessary. Any place they might still be used could simply use AnyStruct without much loss of utility. And I think having an `Integer` type that you can't do integer operations on is likely to cause more confusion that it solves!

Imagine trying to explain to a beginner why this fails... 😅 
```
let a: Integer = 1
let b: Integer = 2

let sum = a + b
```

SupunS:
That's true, this proposal definitely makes the supertypes less useful.
Some scenarios that I had in mind were:
- Interface conformance
- Using in collections. e.g: dictionary key type - where keys are heterogeneous numbers (I agree it's highly unlikely someone would use it though).

Below is a very hypothetical example for interface conformance use-case. Again, I'm not too sure how frequently this pattern is used in the smart contracts in practice, but throwing this just as a 'theoretically' possible/sample use-case :slightly_smiling_face:

```kotlin
pub struct interface Hashable {
  pub fun hash(): Integer  
}

pub struct Foo: Hashable {
  pub fun hash(): Int8 {
      // uses one hashing algirithm
      return 1
  }
}

pub struct Bar: Hashable {
  pub fun hash(): Int64 {
      // uses a different hashing algirithm
      return 1
  }
}
```

turbolent:
Even if supertypes loose some of their usefulness with this proposal, I would still like to go forward with what is currently proposed. We can still propose the removal in a follow-up – that shouldn't influence what is proposed so far

bjartek:
I think this is a very good suggestion. Anything that makes a developer know early that things might not work is an improvement in DevEx if you ask me. +1. 

I can see @dete´s point above though. Very well put.

aishairzay:
+1 great suggestion imo

SupunS:
Thanks, everyone for the feedback! 

I think we can proceed with this FLIP since there aren't any objections against the proposal. The suggested alternative (Dete's suggestion) is an addition to what's proposed here, so we can always do that on top of this proposal.

I will go ahead with the proposal.

bluesign:
@SupunS can you explain me the conformances example, how is it currently and how it will be after this proposal. 

A bit lost on this

```pub struct interface Hashable {
  pub fun hash(): Integer  
}

pub struct Foo: Hashable {
  pub fun hash(): Int8 {
      // uses one hashing algirithm
      return 1
  }
}

pub struct Bar: Hashable {
  pub fun hash(): Int64 {
      // uses a different hashing algirithm
      return 1
  }
}```

SupunS:
@bluesign The conformance example would have no impact. That piece of code currently works and will continue to work even after the change.
I brought up that example as an argument against completely removing number supertypes (to show that it can be useful in some cases).

bluesign:
@SupunS thanks a lot

PR #742, opened 2022-01-03
Cryptography terms improvement

tarakby:
## Description

Small improvement related to the use of cryptography terms. I removed the mention to the signature algorithm ECDSA since it is not relevant in the derivation path case. In order to derive a key from a seed and a path, only the elliptic curve used is needed. The signature algorithm does not impact the derivation result.


psiemens (approved):
LGTM ⭐

JeffreyDoyle (approved):

PR #753, opened 2022-01-11
Variable Transaction Fees: Execution Effort

janezpodhostnik:
## Description

This FLIP builds on the foundations of the [Variable Transaction fees FLIP](20211007-transaction-fees.md) and proposes a model for measuring the execution effort of transactions by choosing certain functions/operations, that are called during the execution of a transaction, to have a related execution effort cost. This FLIP explores a choice of functions/operations and uses data collected from sample transactions and a linear model fit to determine the cost of each chosen functions/operations, so that on average the execution effort of a transaction is proportional to the execution time of the transaction. This FLIP also explores the FLOW cost of a unit of execution effort.


flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,557 @@
...
+
+## Motivation
+
+By improving the model for the execution effort of transactions, transactions will be priced more fairly.
+
+As a consequence transactions that do little (e.g. transferring a (non)fungible token) will cost less, while transactions that do a lot and require a lot of resources to execute will cost more. Attacking the system with heavy transactions in order to overload it will become costlier and thus less viable.
+
+## Current state
+
+Currently execution effort is referenced to as computation cost. It is counted as 1 per every cadence function call or cadence loop made during the transaction.
```
j1010001, line 36:
```suggestion
As of [v0.23.6 release ](https://github.com/onflow/flow-go/tree/v0.23.6), execution effort is referenced to as computation cost. It is counted as 1 per every cadence function call or cadence loop made during the transaction. 
```

j1010001: "currently" is a moving target, better to be more specific :)

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,557 @@
...
+
+The assumption here was that the processing cost of a running a single function <!-- $N$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/4FLaqCTwtI.svg"> times scales linearly with <!-- $N$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/4FLaqCTwtI.svg">. This assumption is made only for transactions where the execution effort of the transaction is not above the execution effort limit.
+
+Good locations and values need to be chosen for the weights so that on average the time taken to execute a transaction (<!-- $t$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/cbdkfkWybi.svg">) will be proportional to the execution effort of that transaction <!-- $\frac{t}{E} = \textbf{const.}$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/ujui9morSm.svg">.
```
j1010001, line 48:
```suggestion
By choosing correct functions for the weights we can find acceptable linear correlation between transaction execution time (<!-- $t$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/cbdkfkWybi.svg">) and it's execution effort <!-- $\frac{t}{E} = \textbf{const.}$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/ujui9morSm.svg">.
```

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,557 @@
...
+    ]
+    ```
+</p>
+</details>
+
+In order to get more varied data, mixed transactions were also simulated. Mixed transactions are created by taking two different sample transactions and running both in one transaction. Each of the sample transaction bodies are run between 1 and _max loop length_/2 times.
+
+### Data Collection procedure
+
+The data was collected by simulating block execution. Each block hah between 1 and 50 transactions (uniformly distributed).
```
j1010001, line 288:
```suggestion
The data was collected by simulating block execution. Each block had between 1 and 50 transactions (uniformly distributed).
```

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,557 @@
...
+| UpdateAccountContractCode  | function calls                           | Updates an account's contract                                                                               |
+| ValueExists                | function calls                           | Checks if a certain register exists                                                                         |
+| ValueDecoded               | function calls                           | Cadence callback after a value was decoded into a cadence type                                              |
+| ValueEncoded               | function calls                           | Cadence callback after a cadence type was encoded into a cadence json                                       |
+
+### Data analysis
+
+The large amount of different weights chosen was intentional. This allows for better insight into which weights should actually be used.
+
+The Following chapters describe how the linear model fitting was done due to a significant amount of outliers in the data and how certain weights were eliminated, because they were either to correlated with other weights, to small, or had a lot of noise.
```
j1010001, line 353:
```suggestion
The Following chapters describe how the linear model fitting was done due to a significant amount of outliers in the data and how certain weights were eliminated, because they were either redundant (correlated with other weights), insignificant (had small or no impact on the correlation), or had a lot of noise.
```

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,557 @@
...
+
+The outliers cause the goodness of the linear model fit to degrade. A way how to remedy this is described in detail in the [Wolfram documentation](https://reference.wolfram.com/applications/eda/RobustFitting.html) and can be described in short as using the residual (error) of each data point to tweak the weight of that data point, and then re-fitting the data with regard to those weights. This is done until the model converges.
+
+The function that converts the residual of a data point to its weight was chosen to be an asymmetrical function with a different cut-off to the left and to the right. This is because the outliers were mostly to the right of the graph, and the signal was to the left of the graph. The cut-off pont was 16 times of the mean of the residuals while the cut-off point to the right was 10 times of the mean residual.
+
+![Residual2Weight](./20220111-execution-effort/residual-to-weight.png)
+
+This improves the final model fit.
```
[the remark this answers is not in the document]
janezpodhostnik: I made this a bit clearer.

pgebheim:
Some high level comments:

1) The meat of this document is describing the methodology for data analysis. The *effects* of this proposal need to be up front, allowing a reader to see: "Before this proposal here are the costs, after this proposal here are the new costs," and this should ideally be in the abstract.

2) This table supposedly summarizes what the FLIP is proposing: https://github.com/onflow/flow/blob/444c9909b1920b579e428eda51e4df2ef142a8fb/flips/20220111-execution-effort.md#final-model-proposal

However, I am not even sure what the units on this chart are and what they mean. Is this e.g. saying that `function_or_loop` currently has a cost of 1 effort, but after this proposal it will have a cost of 0.0141165 effort?

3) The proposal does not discuss how this will be implemented, other than a note of that this can be done without changing the actual fees, but instead just allowing txns to measure execution effort. If that is the case, we need a proposal for what API payloads and data structures need to be maintained in order to support this behavior of reporting execution effort.

4) On the highest level I worry that what we've done here is fragile modeling that points to a conceptually fragile cost framework.

A thousand implementation details could be affecting the various runtime costs of these operations. Future FVM changes could drastically change the actual runtime impact of the various functions, e.g. by adding changing the method of data structure serialization certain operations could become significantly more or less costly.

For an engineer targeting the FVM and actually writing Cadence it is necessary that there is a long term stable *framework* for how they can think of their costs. By attempting to build costs for *high level activities* I think we run the risk of creating a moving target which is either difficult to adjust or difficult to code toward.

Once contracts are in the wild assuming various cost factors (especially related to the max execution cost) it becomes extremely difficult to change these factors. For evidence of this look at the huge conundrum Ethereum went through when attempting to change operation gas costs during the Constantinople hard fork: https://cryptoticker.io/en/ethereum-hard-fork-delayed/

janezpodhostnik:
> 1. The meat of this document is describing the methodology for data analysis. The effects of this proposal need to be up front, allowing a reader to see: "Before this proposal here are the costs, after this proposal here are the new costs," and this should ideally be in the abstract.

The emphasis on a clear description of the process was intentional, as this way it makes the final results transparent and makes the process repeatable which also ties into point 4. I agree that the final results (and the consequences of them) should be better illustrated.

> 2. ... However, I am not even sure what the units on this chart are and what they mean ...

The unit of effort is defined so that the average time taken to execute a transaction is proportional to its effort. This still has a free parameter so a suggested solution is to define that 1 unit of effort corresponds to 1 ms of execution time on an average execution node. This might be glossed over in the FLIP, I will attempt to make this clearer and mark it on the table as well.

> 3. The proposal does not discuss how this will be implemented ...

No, implementation details were left out. It basically just includes keeping a tally of all the weights hit and multiplying that by a few numbers. I will add a short chapter to outline this.

> 4. On the highest level I worry that what we've done here is fragile modelling that points to a conceptually fragile cost framework.

Yes this model is fragile. It is not meant as a final model, but as a first model. Two steps were taken to reduce the fragility:
- only a few weights were chosen this way it is easier to keep track of how many times each transaction hits these specific weights and its execution time and change the weights if needed.
- the process of how the weights are chosen is described in full here and can be repeated.


TODO list for me:
- [ ] describe on a concrete transaction, how the fees will change. That example could be a FLOW transfer
- [ ] update abstract to somehow include the gist of the results
- [ ] more clearly describe the unit of effort
- [ ] add a chapter with a high level of overview of the actual implementation

pgebheim:
> @pgebheim Thank you very much for the input!
> 

Thanks! for the quick reply.

> No, implementation details were left out. It basically just includes keeping a tally of all the weights hit and multiplying that > by a few numbers. I will add a short chapter to outline this.

So I missed the section in #660 which referenced how the fees are emitted via Events. Perhaps just mention that in this doc as well and link to [it in the Variable Transaction Fees flip](https://github.com/onflow/flow/pull/660/files#diff-d4dcee4f8231392b750ad1ef4e848a03483aa455f28153f10e2007b413c97e21R229). I think that describes the public interface changes that would come from this FLIP.

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,555 @@
...
+
+- `ProgramChecked` / `ProgramParsed`
+- `GetCode` / `GetAccountContractCode`
+- `ValueExists` / `CreateAccount`
+
+These pairs of weights were cross checked with the implementation to confirm that they are always called together.
+
+The columns with weights `ProgramChecked`, `GetCode` and `ValueExists` were removed from the weight data matrix <!-- $M$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/SmblEwKco6.svg">.
+
+#### Outliers and robust linear model fitting
```
AlexHentschel, line 373: 😨 I am a bit worried here about the use if robust linear model fitting. As Wolfram explained, robust linear model fitting introduces an axiomatic bias into the model. There is a strong risk here of over-fitting to the training data (benchmark transactions),  so its predictive performance on the training set is better, but potentially worse for unknown transactions. Meaning, the model-generated "execution effort" would not be a good predictor for real-world effort.

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,555 @@
...
+
+The following is a linear model fit on the data and plotting the data on a graph of execution time taken (in milliseconds) vs execution effort.
+
+![DemoFit](./20220111-execution-effort/demo-fit.png)
```
AlexHentschel, line 377: A couple request: 
* could you please include a legend, to clearly identify the `red` as the desired output for a perfect model; and the blue as the predictions of our linear model. 
* Can you reflect the density in your graph. The data points in the around the model completely cover the domain. There could be significant differences in density, which are important to evaluate the performance of our model "by human inspection". 

Lets take a look at the model's performance for fast transactions (short 
To me, it looks there is a very distinct different

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,555 @@
...
+
+The outliers cause the goodness of the linear model fit to degrade. A way how to remedy this is described in detail in the [Wolfram documentation](https://reference.wolfram.com/applications/eda/RobustFitting.html) and can be described in short as using the residual (error) of each data point to tweak the weight of that data point, and then re-fitting the data with regard to those weights. This is done until the model converges. In the wolfram documentation this is called robust fitting.
+
+The function that converts the residual of a data point to its weight was chosen to be an asymmetrical function with a different cut-off to the left and to the right. This is because the outliers were mostly to the right of the graph, and the signal was to the left of the graph. The cut-off point was 16 times of the mean of the residuals while the cut-off point to the right was 10 times of the mean residual.
+
+![Residual2Weight](./20220111-execution-effort/residual-to-weight.png)
```
AlexHentschel, line 385: Where does this graph come from? Sorry, I don't understand. Who determined that this is exactly the shape we wanted to use?

AlexHentschel (commented):
@janezpodhostnik thanks for the detailed write up. I need to understand a bit better how you fitted the model. I am very cautious about using "Robust Fitting". 

My comments below are mainly around consistency of the scientific presentation, highlighting assumptions and notation. Can't say anything about the actual methodology of fitting the model, because there are too many questions for me at the moment.

bluesign:
> The original goal described in the [Variable Transaction fees FLIP](https://github.com/onflow/flow/blob/444c9909b1920b579e428eda51e4df2ef142a8fb/flips/20211007-transaction-fees.md) was that for 95% of transactions to be cheaper after the transition to variable transaction fees.

I think if you have a goal like this, it is also important to put some real cost in the calculation. ( PS: This goal is not realistic but it is not important in this FLIPs context ) 

If we think let's say running an execution node costs X now, out target should be when 100% utilization of execution node, at least `sum(transaction fees) > X `, I think otherwise we are allowing abuse. 

> This means that the 95th percentile execution time times the price of one unit of execution effort should be half of the current fixed fees (the other half will be inclusion effort):

why here is a split of 50/50 ? If there is a calculation here, I think would be nice to share. For me it is not making sense, but maybe I am missing something.

flips/20220111-execution-effort.md
janezpodhostnik: There is a property of the data I tried to include in the model, the "robust fit" + custom weighting function for outliers Is what I used to try to do this.

The property I had in mind (and is also visible in the data) is that for a specific transaction on a specific machine there exists a `t_0` such that the execution time for this transaction on this machine (across multiple runs) is `t=t_min + delta t` where `delta t >= 0`. In other words the machine cannot run the transaction faster than `t_0` but it can (and will) run it slower. (The error of execution time is more positive than negative).

flips/20220111-execution-effort.md
janezpodhostnik: This relates to my reply here https://github.com/onflow/flow/pull/753#discussion_r812110706. I tried to capture a special property of the errors of execution time.

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,555 @@
...
+Matrix <!-- $M$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/SmblEwKco6.svg"> represents the weight data, vector <!-- $v$ --> <img style="transform: translateY(0.1em); background: white;" src="./20220111-execution-effort/eq/LwC51eYP1G.svg"> represents the execution times.
+
+### Weights placement
+
+A total of 26 different weights were used.
+Of those 25 weights were placed in functions in `transactionEnv.go` which implements the interface between Cadence and the flow virtual machine (FVM).
+
+The `function_or_loop_call` weight is an exception as that counts any cadence function (function calls in the cadence script) and any cadence loop. This is also the weight that is already currently in place, as discussed above.
+
+The weights `GetValue` and `SetValue` are also different, as instead of just counting the number of times a transaction calls those functions, they instead count how many bytes were read or written, when they are called.
```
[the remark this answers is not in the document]
janezpodhostnik: Features is a good name. I was also considering just calling them functions, but that seemed too broad. They are more precisely: "a place in the code with an associated weight".

flips/20220111-execution-effort.md
```diff
@@ -0,0 +1,555 @@
...
+
+The aim of this FLIP is to create a model for measuring the execution effort of transactions, that satisfies the following criteria:
+
+- The model must be better then the current model of measuring the execution effort.
+- The model must be straightforward to understand and implement.
+- The implementation of the model must not significantly degrade the execution time of transactions.
+- There must be a clear improvement path for the model.
```
AlexHentschel, line 22: I think there is one very important requirement missing:
> * The model must be byzantine compatible, i.e. hard to game / circumvent. 

This is probably a criterium that is hard to completely enforce, but I think we need a reasonable level of confidence here.

pgebheim:
Merging this pull request so that the canonical document can be linked. Any future edits should come in the form of a secondary PR.

PR #764, opened 2022-01-18
Fix relative links pt2.

10thfloor:
Closes: #736 

## Description

Content on the docs site, pulled from github – links in those must be modified to account for the difference in relative depth between the docs site and the contining repo.

Example, a document in Github, with a relative link; `../install.md` must be transformed to `../../install` in order to work with the docs site content hierarchy. 

Details are discussed in the issue.


docs/content/node-operation/index.mdx
```diff
@@ -13,3 +13,5 @@ To set up a new Flow node you will need to complete the following steps:
 3. [Start](/node-operation/node-bootstrap#step-2---getting-ready-to-start-your-flow-node) your node!
 
 
+[test](node-roles.mdx)
```
10thfloor, line 16:
```suggestion
```

kerrywei: is this test link intended?

kerrywei: is the test link intended?

docs/plugins/gatsby-remark-relative-linker/functions.js
```diff
@@ -0,0 +1,59 @@
+const sibling = "./";
+const parent = "../";
+const UNPREFIXED = -1;
+const wasSeen = {};
```
kerrywei, line 4: this const is not used

docs/plugins/gatsby-remark-relative-linker/functions.js
```diff
@@ -0,0 +1,59 @@
+const sibling = "./";
+const parent = "../";
```
kerrywei, line 2: shall we make all const upper cases (or all lower cases -- to follow whatever the convention is in JS)

docs/plugins/gatsby-remark-relative-linker/functions.js
```diff
@@ -0,0 +1,59 @@
...
+
+  if (result) {
+    return updateLinkText(linkText, forDepth, isIndex);
+  }
+}
+
+function updateLinkText(linkText, depth, index) {
+  switch (index) {
+    case true:
+      return depth === UNPREFIXED || 0 ? sibling + linkText : linkText;
```
kerrywei, line 38: curious: what's the difference between ./a.md and a.md? i.e. does it make a difference in terms of interpreting the URL?

docs/plugins/gatsby-remark-relative-linker/functions.js
```diff
@@ -0,0 +1,59 @@
...
+    case false:
+      return depth === UNPREFIXED || 0
+        ? parent + linkText
+        : linkText.replace(
+            buildDepthCheckRegex(depth),
+            parent.repeat(depth + 1)
+          );
```
kerrywei, line 45: in the case false scenario, would 'parent + linkText' to be good enough (always pre-fix ../) ?

docs/plugins/gatsby-remark-relative-linker/functions.js
10thfloor: Just making sure all sibling links end up with the `./` prefix, for consistency.

docs/plugins/gatsby-remark-relative-linker/functions.js
10thfloor: Yes, all sibling links on a non-index page need the `../` prefix.

docs/plugins/gatsby-remark-relative-linker/functions.js
10thfloor, line 4:
```suggestion
```

docs/plugins/gatsby-remark-relative-linker/index.js
```diff
@@ -1,28 +1,19 @@
...
-        node.url = node.url.replace(/^\.\.\/\.\.\//, "../../../");
-      }
+      const isIndex =
+        markdownNode.fileAbsolutePath.includes("index.md") ||
+        markdownNode.fileAbsolutePath.includes("README.md");
 
-      if (foundNoPrefix) {
-        node.url = "../" + node.url;
-      }
+      node.url = updateLinkText(node.url, isIndex);
```
kerrywei, line 16: would this work if the function signature does not match? if it does not work, this indicates the necessity of including docs/plugins/gatsby-remark-relative-linker/test.js to be run in CI so that we can capture this during code review

docs/plugins/gatsby-remark-relative-linker/functions.js
```diff
@@ -0,0 +1,60 @@
...
+      return depth === UNPREFIXED ? SIBLING + linkText : linkText;
+    case false:
+      return depth === UNPREFIXED
+        ? PARENT + linkText
+        : depth === 0
+        ? linkText.replace(SIBLING, PARENT)
+        : linkText.replace(
+            buildDepthCheckRegex(depth),
+            PARENT.repeat(depth + 1)
+          );
```
kerrywei, line 46: it might be just me but I find the nested ? : syntax quite not straightforward to follow. Originally I was thinking this function can be simplified to be: 

function updateLinkText(linkText, depth, index) {
  switch (index) {
    case true:
      return linkText;
    case false:
      return "../" + linkText;
}

Another reason behind prefixing ../ directly is: 
what if developer writes language/../../docs/introduction.md -- would this work with current logic? I think this change will not process this case as it does not begin with ../../ ?

docs/plugins/gatsby-remark-relative-linker/functions.js
```diff
@@ -0,0 +1,53 @@
...
+  for (const depth of depthCheck(MAX_DEPTH_CHECK)) {
+    const updatedLinkText = doUpdateDepth(linkText, depth, isIndex);
+    if (updatedLinkText) {
+      return updatedLinkText;
+    }
+  }
+  return linkText;
+}
+
+module.exports = updateRelativeDepth;
```
kerrywei, line 53: I think this PR will work, but curious could this version also work? Basically I'm trying to understand the necessity of having the for-loop in line 44: is it true that regardless of the depth we want to check, as long as the link is inside a page that is NOT index.md or README.md, prefixing ../ would be all we need? Let me know if I missed anything 

```
const PARENT = "../";

function updateRelativeDepth(linkText, index) {
  switch (index) {
    case true:
      return linkText;
    case false:
      return PARENT + linkText;
  }
}

module.exports = updateRelativeDepth;
```

10thfloor: lol yup, it looks like that will work. Removing all the code now.

kerrywei (approved):
thank you for addressing this

PR #788, opened 2022-01-31
[Flips] Introducing Soul-Bound Token (SBT) to cadence and flow exosystem

Bamthelearner:
This proposal is to introduce a standard for Soul-Bound Token to the flow ecosystem.
Soul-Bound Token means tokens that are Non-Fungible yet not transferrable to the others.
Soul-Bound Token (SBT) should be differentiated from NFTs for long-run to faciliate easy
integration of SBTs.


flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+        pub fun deposit(token: @SBT)
+
+        // getIDs returns an array of the IDs that are in the collection
+        pub fun getIDs(): [UInt64]
+
+        // Returns a borrowed reference to a SBT in the collection
+        // so that the caller can read data and call methods from it
+        pub fun borrowSBT(id: UInt64): &SBT {
+            pre {
+                self.ownedSBTs[id] != nil: "SBT does not exist in the collection!"
```
scotter567, line 108: Question: If the index of some owdeSBT is not nil wouldn't that imply that there they do own the SBT? not sure If i just don't understand how borrowing works here.

Bamthelearner: in cadence, the arguement on precondition should be true in order not to return the warning that comes after it.
So in this case, it should be :
The ownedSBTs[id] should not be nil (true)
if it is nil (false), then panic("SBT does not exist in the collection")

bluesign:
I think we need different collection for SBTs, instead of each SBT having its own collection. 

So NFT contract as a template can be not the best idea. 

I see different parties minting SBTs: 

TSSBT, allDaySBT, FLOAT etc but I think users would like to store them in single collection. 

Otherwise I can sell my TSSBT collection, and forcing soulblund can be tricky.

joshuahannan:
Yeah, if we truly want to have a type of token that can't be transferred between accounts, then we'd have to implement some sort of central ledger, or introduce a new feature to Cadence that prevents transfers

LanfordCai:
I have a question: How to define soul-bound in Flow/Cadence?
Token can be transferred, and also, Collection can be transferred, and even the Account itself can be swapped by replacing keys.

bluesign:
@LanfordCai I think account = identity is good starting point. I mean we can skip account swap scenario. Collection can be swapped, but we can make collection to work only when it is stored in `one specific account`.

flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+
+        // deposit takes a SBT as an argument and adds it to the Collection
+        //
+        pub fun deposit(token: @SBT)
+    }
+
+    // Interface that an account would commonly 
+    // publish for their collection
+    pub resource interface CollectionPublic {
+        pub fun deposit(token: @SBT)
```
KhaledAylii, line 84: We would want to be able to extend this by importing the contracts interfaces
```suggestion
        pub fun deposit(token: @AnyResource{ISBT})
```

flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+        pub fun getIDs(): [UInt64]
+        pub fun borrowSBT(id: UInt64): &SBT
```
KhaledAylii, line 86:
```suggestion
        pub fun borrowSBT(id: UInt64): &AnyResource{ISBT}
```

flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+    }
```
KhaledAylii, line 87: I recommend we also have a function to destroy an SBT
```suggestion
        destroySBT(id: UInt64)
    }
```

flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+        // deposit takes a SBT as an argument and adds it to the Collection
+        //
+        pub fun deposit(token: @SBT)
+    }
+
+    // Interface that an account would commonly 
+    // publish for their collection
+    pub resource interface CollectionPublic {
+        pub fun deposit(token: @SBT)
+        pub fun getIDs(): [UInt64]
```
KhaledAylii, line 85: I think we should still have a withdraw function so that if a collection includes both SBT and regular NFTs, we insist that there's the following precondition
```suggestion
        pub fun getIDs(): [UInt64]
        pub fun withdraw(id: UInt64) {
           post {
              result.isInstance(Type<@AnyResource{ISBT}>()): "withdrawing an SBT is forbidden"
           }
         }
        
```

flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+    // to be declared in the implementing contract
+    //
+    pub resource Collection: Receiver, CollectionPublic {
+
+        // Dictionary to hold the SBTs in the Collection
+        pub var ownedSBTs: @{UInt64: SBT}
+
+        // deposit takes a SBT and adds it to the collections dictionary
+        // and adds the ID to the id array
+        pub fun deposit(token: @SBT)
```
KhaledAylii, line 99: is this meant to be unimplemented in the final pull request?

flips/20220131-sbt-integration/SoulBoundToken.cdc
```diff
@@ -0,0 +1,120 @@
...
+
+        // getIDs returns an array of the IDs that are in the collection
+        pub fun getIDs(): [UInt64]
```
KhaledAylii, line 102: same here?

KhaledAylii (commented):
Few considerations and suggestions - I think this is a fantastic idea, the main idea would be to allow deposit, disallow withdraw if the NFT implements `ISBT`. I notice i added some of the comments in the public interface, just assume I meant the actual resource

pgebheim:
Closing as abandoned

PR #790, opened 2022-01-31
removing the signer_index field from the transaction signature in the…

vishalchangrani:
… open API spec

Closes: #???

## Description

______

For contributor use:

### DIFF
--- a/openapi/go-client-generated/api/swagger.yaml
+++ b/openapi/go-client-generated/api/swagger.yaml
@@ -809,11 +809,9 @@ components:
           - address: null
             signature: ""
             key_index: key_index
-            signer_index: signer_index
           - address: null
             signature: ""
             key_index: key_index
-            signer_index: signer_index
           arguments:
           - ""
           - ""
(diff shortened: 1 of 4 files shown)


openapi/go-client-generated/api/swagger.yaml
```diff
@@ -809,11 +809,9 @@ components:
           - address: null
             signature: ""
             key_index: key_index
-            signer_index: signer_index
```
vishalchangrani, line 812: The `signer_index` field is populated at the backed and should not be passed in from the client. The gRPC spec does not defines it as part of the transaction signature - https://github.com/onflow/flow/blob/master/protobuf/flow/entities/transaction.proto#L24-L28

I am not aware of a use case where the signer index should be passed in by the REST client. Passing in the signer_index currently causes the transaction ID (hash of the transaction fields) to be different between the Access node and the Collection node since the collection node [calculates the signer_index](https://github.com/onflow/flow-go/blob/bd15c090c932cebbfeacbcf0c1a0bb10019b0f1d/model/flow/transaction.go#L314-#L318)  while the Access node uses what came in the request..

smnzhu (approved):

PR #809, opened 2022-02-08
Reorganise sporks.json for easier querying

m4ksio:
Changes sporks.json structure to
a) indicate is spork is live
b) add seed nodes with keys for live sporks
c) add multiple ANs
d) change structure so it can be queried easily (no need to search arrays for items).


smnzhu (commented):
Hey @m4ksio, do you have a tool you use to generate the sporks.json? If you do, I think it would definitely be good to have that in version control somewhere, and if not, we should create a tool because with such a large file like this it's extremely easy to make manual errors.

Secondly, should we have a separate file per spork to retain the history of sporks instead of replacing the file every time?

m4ksio:
> Hey @m4ksio, do you have a tool you use to generate the sporks.json? If you do, I think it would definitely be good to have that in version control somewhere, and if not, we should create a tool because with such a large file like this it's extremely easy to make manual errors.

We don't have tools. I would imagine this will follow once we establish some standard.

> Secondly, should we have a separate file per spork to retain the history of sporks instead of replacing the file every time?

This is still an open question really, so far I figured its easier to have one file which you can check for content and new sporks rather than check a files which may or may not exist.

smnzhu:
@m4ksio great, thanks :)

Created an issue just to make sure we track this: https://github.com/onflow/flow/issues/814

PR #834, opened 2022-02-16
Added missing doc updates for ExecutionDataID

franklywatson:
We found that ExecutionDataID was newly in the code but not updated in the docs. This fixes that omission

### DIFF
--- a/docs/content/access-api.md
+++ b/docs/content/access-api.md
@@ -1095,6 +1095,7 @@ message ExecutionResult {
   bytes block_id
   repeated Chunk chunks
   repeated ServiceEvent service_events
+  bytes execution_data_id  
 }
 ```
 
@@ -1104,6 +1105,7 @@ message ExecutionResult {
 | block_id           | ID of the block this execution result corresponds to |
 | chunks             | Zero or more chunks                                  |
 | service_events     | Zero or more service events                          |
+| execution_data_id  | Root ID of the Execution Data blob tree              |
 
 
 ### Chunk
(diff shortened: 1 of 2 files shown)


docs/content/access-api.md
```diff
@@ -1095,6 +1095,7 @@ message ExecutionResult {
   bytes block_id
   repeated Chunk chunks
   repeated ServiceEvent service_events
+  bytes execution_data_id
```
vishalchangrani, line 1098: the change for this hasn't yet been deployed. So technically this field will not be returned yet. 
Can we queue this up after this PR https://github.com/onflow/flow-go/pull/2014 has been merged and deployed?

openapi/go-client-generated/docs/ExecutionResult.md
```diff
@@ -7,6 +7,7 @@ Name | Type | Description | Notes
 **BlockId** | **string** |  | [default to null]
 **Events** | [**[]Event**](Event.md) |  | [default to null]
 **Links** | [***Links**](Links.md) |  | [optional] [default to null]
+**ExecutionDataID** | **string**  |  | [default to null]
```
smnzhu, line 10: It looks like this is an auto-generated file. You should probably change the definition at the source (ie. the OpenAPI spec), and then regenerate this file.

However, currently we haven't updated the REST API to have this change, so for now we can probably omit this entirely.

docs/content/access-api.md
franklywatson: Yes of course!

openapi/go-client-generated/docs/ExecutionResult.md
franklywatson: ok will do

turbolent:
Sorry, closed this accidentally

PR #839, opened 2022-02-22
adding a page listing current gRPC Access API rate limits

vishalchangrani:
Closes: https://github.com/onflow/flow/pull/839/

## Description

______

For contributor use:


docs/content/access-api-rate-limits.mdx
```diff
@@ -0,0 +1,64 @@
...
+Once the limit has reached, the client will receive a rpc error 1ResourceExhausted` in the gRPC response.
+
+Pleqse note, these limits only apply to the access nodes hosted by DapperLabs. Access node run by other node operators will have different rate limits.
+
+
+##### Mainnet
+
+|                 gRPC API                             | Total request per second per IP               |
+|:-----------------------------------------------------|:---------------------------------------------:|
+| flow.access.AccessAPI/Ping                           |                   2000                        |
```
vishalchangrani, line 17: we have 4 public facing ANs with the limit set to 500 on each.

docs/content/access-api-rate-limits.mdx
```diff
@@ -0,0 +1,64 @@
+---
+title: Flow Access API Rate limits on DapperLabs Access nodes
+---
+
+#### Rate limits
+
+Following are the current rate limits for the [Access Node gRPC API](/access-api).
+Once the limit has reached, the client will receive a rpc error 1ResourceExhausted` in the gRPC response.
+
+Pleqse note, these limits only apply to the access nodes hosted by DapperLabs. Access node run by other node operators will have different rate limits.
```
turbolent, line 10:
```suggestion
title: Flow Access API rate limits on Dapper Labs Access nodes
---

#### Rate limits

Following are the current rate limits for the [Access Node gRPC API](/access-api).
Once the limit has reached, the client will receive an RPC error `ResourceExhausted` in the gRPC response.

Please note, these limits only apply to the access nodes hosted by Dapper Labs. Access nodes run by other node operators will have different rate limits.
```

docs/content/access-api.md
```diff
@@ -25,6 +25,9 @@ The Access Nodes hosted by DapperLabs are accessible at:
 
 `access.canary.nodes.onflow.org:9000`
 
+#### Ratelimits for DL access nodes
+[Rate limit](/access-api-rate-limits)
```
turbolent, line 29:
```suggestion
#### Rate limits for Dapper Labs access nodes

Access nodes operated by Dapper Labs are [rate limited](/access-api-rate-limits).
```

docs/content/access-api-rate-limits.mdx
```diff
@@ -0,0 +1,64 @@
...
+Once the limit has reached, the client will receive an RPC error `ResourceExhausted` in the gRPC response.
+
+Please note, these limits only apply to the access nodes hosted by Dapper Labs. Access nodes run by other node operators will have different rate limits.
+
+
+##### Mainnet
+
+|                 gRPC API                             | Total request per second per client IP        |
+|:-----------------------------------------------------|:---------------------------------------------:|
+| flow.access.AccessAPI/Ping                           |                   2000                        |
```
turbolent, line 17: Given that the whole document is for access node APIs, maybe remove the `flow.access.AccessAPI/` prefix, it is always the same and implied

docs/content/access-api-rate-limits.mdx
peterargue: It's probably worth adding a note about that above the table so people don't get confused if they send requests to an individual node.

The limits are the same for gRPC-Web as well right?

vishalchangrani: added note abt both those things.

docs/content/access-api-rate-limits.mdx
vishalchangrani: so i wanted to differentiate this with the legacy API which I think has a different prefix.

sideninja (approved):

docs/content/access-api-rate-limits.mdx
peterargue: any idea if they also use the same counters

docs/content/access-api-rate-limits.mdx
```diff
@@ -0,0 +1,66 @@
+---
+title: Flow Access API rate limits on Dapper Labs Access nodes
+---
+
+#### Rate limits
+
+Following are the current rate limits for the [Access Node gRPC API](/access-api) in total across all Dapper Labs Access nodes.
```
peterargue, line 7: Since the load balancer is operating on requests from all clients, if 2 clients are sending requests at the same time, it's possible their requests are not evenly distributed. Worst case, all requests could go to a single node and they hit the rate limit after 500 requests, and best case they are evenly distributed and hit it after 2000. I think in practice, users will run into some variation is the effective limits, so I tried to clarify that below.

```suggestion
Following are the current rate limits for the [Access Node gRPC API](/access-api) in total across all Dapper Labs Access nodes. The actual limits are enforced by each individual node. If you are using a load balanced endpoint, the limits below are the max rate, and may vary depending on how your requests are routed.
```

vishalchangrani: make sense

peterargue (approved):

PR #841, opened 2022-02-23
Adding missing fields to block header proto for correct hash validation

koko1123:
## Description
Add remaining fields to protobuf entity to match `header.go`


franklywatson:
There are no tests that go with this? Maybe a concern? Should make sure the coverage is where it should be when we follow up with @sideninja's suggestions

koko1123:
> There are no tests that go with this? Maybe a concern? Should make sure the coverage is where it should be when we follow up with @sideninja's suggestions

No tests in the `flow` repo as it just handles entity/openAPI definitions, will have corresponding tests in `flow-go` a la: https://github.com/onflow/flow-go/pull/2006

smnzhu (approved):

peterargue (approved):

PR #843, opened 2022-02-23
Add Protoc fields for requested responses (block data) 

koko1123:
fixes #2044, #2049, #2051
Setup protocolbuffer entitites to have the new fields without breaking backward compatibility


protobuf/flow/entities/block_header.proto
```diff
@@ -9,13 +9,14 @@ import "google/protobuf/timestamp.proto";
...
-  bytes parent_id = 2;
-  uint64 height = 3;
-  bytes payload_hash = 4;
-  google.protobuf.Timestamp timestamp = 5;
-  uint64 view = 6;
-  repeated bytes parent_voter_ids = 7;
-  bytes parent_voter_sig_data = 8;
-  bytes proposer_id = 9;
-  bytes proposer_sig_data = 10;
+  bytes chain_id = 2;
```
m4ksio, line 12: I think changing field order will break backwards compatibility.

smnzhu (changes requested):
Your previous PR changed the order of the timestamp field: https://github.com/onflow/flow/pull/841/files

Can you make sure you change it back in this PR?

smnzhu:
And going forward for the rest of the PRs, please just remember not to change the order of the existing fields. Thanks :)

protobuf/flow/entities/block_header.proto
```diff
@@ -11,11 +11,12 @@ message BlockHeader {
   bytes id = 1;
   bytes parent_id = 2;
   uint64 height = 3;
-  bytes payload_hash = 4;
-  google.protobuf.Timestamp timestamp = 5;
+  google.protobuf.Timestamp timestamp = 4;
+  bytes payload_hash = 5;
```
tav, line 15: With protobuf, one shouldn't change the field numbers for any existing fields as it would break old clients.

See the [Updating a Message Type](https://developers.google.com/protocol-buffers/docs/proto#updating) documentation for more info.

tav (changes requested):

protobuf/flow/entities/block_header.proto
smnzhu: @tav Thanks, as I mentioned in my comment above: https://github.com/onflow/flow/pull/843#pullrequestreview-891943697

This is actually a change that was mistakenly made a previous PR, and this PR is reverting the order change.

smnzhu (approved):

PR #853, opened 2022-03-01
Adjust chainID type to native string

koko1123:
small miss, `String([]byte)` cast doesn't compensate for the forward conversion


m4ksio:
Just to confirm - this field has been added very recently in other PRs, right?

koko1123:
> Just to confirm - this field has been added very recently in other PRs, right?

Yes, no backward compatibility issue sisnce it isn't deployed

smnzhu (approved):

PR #856, opened 2022-03-02
Add and Generate pb fields to leave deprecated use fields from previous structs in collection_guarantee, block_seal and  block_header

koko1123:
Generated pb file for this issue: https://github.com/onflow/flow/pull/853


smnzhu (approved):

zhangchiqing (approved):

protobuf/flow/entities/block_seal.proto
```diff
@@ -11,4 +11,11 @@ message BlockSeal {
   repeated bytes execution_receipt_signatures = 3;
   repeated bytes result_approval_signatures = 4;
   bytes final_state = 5;
+  bytes result_id = 6;
```
zhangchiqing, line 14: I guess we append fields with increasing number (`6`) is so that it's backward compatible, right?

If we are not going to remove the deprecated fields, it's better to add some comments for those fields:
```
  bytes execution_receipt_id = 2;  // deprecated field
  repeated bytes execution_receipt_signatures = 3;  // deprecated field
  repeated bytes result_approval_signatures = 4;  // deprecated field
```

PR #859, opened 2022-03-03
Use proper seed nodes ports

m4ksio:


peterargue (approved):
good catch

sporks.json
```diff
@@ -33,11 +33,11 @@
         },
         "seedNodes": [
           {
-            "address": "access-007.mainnet16.nodes.onflow.org:3569",
+            "address": "access-007.mainnet16.nodes.onflow.org:3570",
```
zhangchiqing, line 36: why? I have no context about it

m4ksio: We moved separate public p2p network some months ago, hence we use separate ports.

vishalchangrani (approved):

PR #869, opened 2022-03-10
updating the migration guide

vishalchangrani:
Closes: #???

## Description

______

For contributor use:

### DIFF
--- a/docs/content/node-operation/node-migration.mdx
+++ b/docs/content/node-operation/node-migration.mdx
@@ -11,13 +11,17 @@ Choose the method depending upon what part of the staking data of the node is ch
 
 If there is no change to the network address or the staking and networking keys and only the hardware the node is running needs to be changed then do the following:
 1. Stop the Flow node.
-2. Copy over the bootstrap data (typically under `/var/flow/data`) which contains the node private key to the new machine.
-3. Make sure the new node will run on the same network address as the old one.
+2. Copy over the bootstrap data (typically under `/var/flow/bootstrap`) which contains the node private key to the new machine.
+3. Copy over the data folder (typically under `/var/flow/data`) which contains the state data.
+4. Start the new node on the same network address as the old one.
+
+<Callout type="warning">
 Please ensure that there is minimal downtime during this migration.
+</Callout>
 
 <Callout type="warning">
 The network address is currently part of the staking data that was submitted for the node. It is how other nodes in the network discover this node.
-Hence, the network address of the node must stay the same between sporks otherwise the node will become unreachable for the other nodes and stop functioning.
+Hence, the network address of the node must stay the same between epochs otherwise the node will become unreachable for the other nodes and stop functioning.
 </Callout>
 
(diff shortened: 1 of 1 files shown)


docs/content/node-operation/node-migration.mdx
```diff
@@ -11,21 +11,25 @@ Choose the method depending upon what part of the staking data of the node is ch
...
 
 <Callout type="warning">
 The network address is currently part of the staking data that was submitted for the node. It is how other nodes in the network discover this node.
-Hence, the network address of the node must stay the same between sporks otherwise the node will become unreachable for the other nodes and stop functioning.
+Hence, the network address of the node must stay the same between epochs otherwise the node will become unreachable for the other nodes and stop functioning.
 </Callout>
 
 ### Method 2 - Network address change
 
 A change to the node network address (IP or a hostname) can only be done during the spork process.
```
vishalchangrani, line 29: The nodes retain their pub-sub connections before and after an epoch transition. Hence, I am not sure what will happen if the network address is changed at an epoch boundary (before a spork) and the nodes are not restarted. If the node whose address is being changed itself is restarted then it may work since it will force the other nodes to pick up the new address but the node has to be restarted after the epoch transition.

jordanschalm: Changing network addresses was never intended for use between epochs. It wasn't implemented with that in mind and it wasn't tested. I think we should continue communicating that network address changes can only happen during sporks. If there is enough demand for per-epoch changes, we should revisit the implementation with that in mind and test that flow properly.

docs/content/node-operation/node-migration.mdx
```diff
@@ -11,21 +11,25 @@ Choose the method depending upon what part of the staking data of the node is ch
...
 
 To change the networking address:
-1. A day before the [upcoming mainnet spork](/node-operation/upcoming-sporks.mdx), change the network address for the nodes in Flow Port (using the update network address feature).
+1. A day before the [upcoming mainnet spork](/node-operation/upcoming-sporks), change the network address for the nodes in Flow Port (using the update network address feature).
 The change will not take effect till an epoch transition happens.
```
jordanschalm, line 33: It should take effect immediately after the spork, right? Since we will be pulling down node info from the chain now.

jordanschalm (approved):

docs/content/node-operation/node-migration.mdx
vishalchangrani: agree 100%

docs/content/node-operation/node-migration.mdx
vishalchangrani: yeah thats why I mention on line 31 below that spork causes the epoch transition.

vishalchangrani: will that suffice?

jordanschalm: Oh I missed that since it wasn't in the diff. I think that works 👍

PR #874, opened 2022-03-14
add optional flag for full block responses

koko1123:
Add flag in block and block header requests to toggle between full and light response types


protobuf/flow/access/access.proto
```diff
@@ -128,32 +128,39 @@ message PingResponse {}
...
 
 message GetBlockHeaderByHeightRequest {
   uint64 height = 1;
+  optional bool full_block_header = 2;
 }
 
 message BlockHeaderResponse {
   entities.BlockHeader block = 1;
+  optional bool full_block_header = 2;
 }
```
smnzhu, line 147: Do we actually need this for block header? I thought the size increase was only relevant for the full block?

protobuf/flow/access/access.proto
```diff
@@ -128,32 +128,39 @@ message PingResponse {}
...
 
 message GetBlockHeaderByHeightRequest {
   uint64 height = 1;
+  optional bool full_block_header = 2;
 }
 
 message BlockHeaderResponse {
   entities.BlockHeader block = 1;
+  optional bool full_block_header = 2;
```
smnzhu, line 146: probably not necessary for the response

protobuf/flow/access/access.proto
koko1123: I did it to cut down on `ParentVoterIDs` and `ParentVoterSigData` that used to just be a composite object

koko1123: actually nver mind, you're right, even the signer ID size doesn't baloon too much, and we don't include the block heder in the light version of the `block` response anyway

smnzhu (approved):

PR #875, opened 2022-03-15
Create RPC endpoints for GetTransactionResultByIndex in AN and EN

koko1123:
Closes: #2100
https://github.com/onflow/flow-go/issues/2100

create endpoints in AN for `GetTransactionResultByIndex` and matching endpoint in EN so the request can be forwarded


protobuf/flow/access/access.proto
```diff
@@ -187,6 +190,11 @@ message GetTransactionRequest {
   bytes id = 1;
 }
 
+message GetTransactionByIndexRequest {
+  bytes id = 1;
```
smnzhu, line 194: nit: block_id, to be completely clear.

smnzhu (approved):

PR #881, opened 2022-03-23
FLIP Contract Import Syntax

sideninja:
## Description
This is a FLIP PR proposing a new cadence contract import syntax.
______

For contributor use:


bluesign:
Few related questions/ideas mostly related to cadence

- versioning of contracts (import specific version of the contract) I think this is a must have for compose-ability. But also little risky/tricky

-  dynamic import: ( import from variable to a variable )

- name service: this would be nice to have but also competing with existing community projects

sideninja:
> versioning of contracts (import specific version of the contract) I think this is a must have for compose-ability. But also little risky/tricky

Yeah, I think there are a bunch of problems still needing to be solved but the nice thing about it is that we can do that as we learn more about it, it's not restricting any syntax at this point. Versioning is something that at this point I wouldn't like to solve as you mentioned is still a bit tricky and it might require some other solutions to come along. I don't want to do too much with this proposal so it's actually doable soon. 

> dynamic import: ( import from variable to a variable )
Good feature, but again maybe we can add that later if needed.

> name service: this would be nice to have but also competing with existing community projects 
Yes again this should just first be established and then incorporated in the syntax. That is the reason I'm not restricting the syntax too much yet.

Thank you for your input :)

PR #892, opened 2022-03-31
FLIP: Contract Import Improvements

sideninja:
This is a FLIP PR proposing a new cadence contract import syntax.
______

For contributor use:

### DIFF
--- a/flips/2022-03-23-contract-imports-syntax.md
+++ b/flips/2022-03-23-contract-imports-syntax.md
@@ -21,7 +21,7 @@ switching betwen those environments hard.
 
 ### Declaring an Import
 
-There are currently two import replacement mechanisms used in cadence contracts:
+There are currently two import replacement mechanisms used in Cadence contracts:
 
 #### **Imports referencing a placeholder**
 
--- a/flips/2022-03-23-contract-imports-syntax.md
+++ b/flips/2022-03-23-contract-imports-syntax.md
@@ -10,7 +10,7 @@
 ## Objective
 Uniforming the import syntaxes that are used in Cadence contracts. Currently there are few established patterns in the community for how to 
 write imports that would then be replaced before deploying the code to the network. The replacement mechanism is needed because depending 
-on the network you develop on the addresses will change and in some cases it might even reference a local file.
+on the network you develop on, the addresses will change, and in some cases it might even reference a local file.
 
 ## Motivation
 Currently there is no single standard for how to solve this problem and hence few have been established which makes it super hard 
--- a/flips/2022-03-23-contract-imports-syntax.md
+++ b/flips/2022-03-23-contract-imports-syntax.md
@@ -2,7 +2,7 @@
 
 | Status        | Proposed       |
 :-------------- |:---------------------------------------------------- |
-| **FLIP #**    | |
+| **FLIP #**    | 892 |
 | **Author(s)** | Gregor Gololicic @sideninja gregor.gololicic@dapperlabs.com |
 | **Sponsor**   | Gregor Gololicic @sideninja gregor.gololicic@dapperlabs.com |
 | **Updated**   | 2022-03-23                                           |
--- a/flips/2022-03-23-contract-imports-syntax.md
+++ b/flips/2022-03-23-contract-imports-syntax.md
@@ -123,6 +123,9 @@ Frontend applications could fetch the configuration as a resource from the webse
 
 Backend applications should import the configuration and replace the imports based on the environment in which the backend application is being run.
 
+**Cadence**
+There is no change needed in the Cadence implementation as it already supports string locations (i.e. the syntax import "contract").
+
 ### Future-Proofing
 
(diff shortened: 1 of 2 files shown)


sideninja:
Removed PR by mistake so I'm pasting this from the previous PR https://github.com/onflow/flow/pull/881

@bluesign

Few related questions/ideas mostly related to cadence
- versioning of contracts (import specific version of the contract) I think this is a must have for compose-ability. But also little risky/tricky
- dynamic import: ( import from variable to a variable )
- name service: this would be nice to have but also competing with existing community projects

Response:
versioning of contracts (import specific version of the contract) I think this is a must have for compose-ability. But also little risky/tricky

Yeah, I think there are a bunch of problems still needing to be solved but the nice thing about it is that we can do that as we learn more about it, it's not restricting any syntax at this point. Versioning is something that at this point I wouldn't like to solve as you mentioned is still a bit tricky and it might require some other solutions to come along. I don't want to do too much with this proposal so it's actually doable soon.

dynamic import: ( import from variable to a variable )
Good feature, but again maybe we can add that later if needed.

name service: this would be nice to have but also competing with existing community projects
Yes again this should just first be established and then incorporated in the syntax. That is the reason I'm not restricting the syntax too much yet.

Thank you for your input :)

flips/2022-03-23-contract-imports-syntax.md
```diff
@@ -0,0 +1,150 @@
...
+}
+...
+}
+```
+
+Another benefit to this file is to have a way to define mock contracts for testing. You wouldn't be limited to defining one address per network but you could define multiple environments.
+
+The above example is just an idea for values this file needs to include, the file format is up for discussion, but we could argue `flow.json` might be a good place to put this information especially since it can be used that it just defines this section and nothing else.
+
+### Implementation
```
turbolent, line 116: Maybe note here that there is no implementation needed in Cadence itself, as it already supports string locations (i.e. the syntax `import "contract"`)

turbolent (approved):
Nice, this is going to make tooling much better!

flips/2022-03-23-contract-imports-syntax.md
sideninja: good flag thanks

flips/2022-03-23-contract-imports-syntax.md
```diff
@@ -0,0 +1,150 @@
...
+### Current Implementation Problems
+
+**Imports referencing a relative path**
+
+- there will not always be a contract present locally (or at least it shouldn't be as the network is the source of truth for a contract and downloading the contract has definite downsides especially contracts that are still able to update)
+- potential problems across different file systems
+
+**Imports referencing a placeholder**
+
+- Not a valid cadence syntax
```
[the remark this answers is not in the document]
sideninja: so disrespectful uff :D

flips/2022-03-23-contract-imports-syntax.md
```diff
@@ -0,0 +1,153 @@
...
+
+`import <declarations to import> from "<namespaced location>"`
+
+Example:
+
+`import NonFungibleToken from "onflow/NonFungibleToken"`
+
+The source files should be accompanied by a configuration/dependency resolution file, defining locations of contracts addresses on a specific network. Including this file is beneficial so developers exploring contract source files in public repositories can understand which contracts this Cadence code imports. Think of this as go.mod, package.json or any other implementation of this idea where source code that imports something needs to define what that is in the supporting "configuration" file.
+
+Example flow.json for the above syntax:
```
turbolent, line 94: As discussed, maybe point out this is only an example how a tool like the CLI may use the syntax proposed this standard, and the behaviour of the tools is not part of this proposal

bjartek:
Have you consider supporting aliasing? The name of a contract is not unique across the chain and in the future I am pretty sure there will be more contracts that are called common names. Right now it is not possible to import two `Profile` contracts from two different addresses. 

So something like

`import Profile from find/Profile as FindProfile`
`import Profile from qvvg/Profile as QvvgProfile`

for instance. 

Also have you considered the short for of just using
`import NonFungibleToken`

Flow.json will have the inforamtion about that if it is not namespace. So if the contract part of flow.json contains

```
"conctracts" : {
  "NonFungibleToken" : "cadence/contracts/NonFungibleToken"
}
```

sideninja:
I see @bjartek that's a valid flag for sure, the only thing I'm not quite sure about is that it feels to me like this could/should be supported by Cadence natively. What do you think also cc @turbolent

bjartek:
I guess the core need here is the ability to migrate state to a new version of a contract.

turbolent:
Support for aliasing in imports has been requested before in https://github.com/onflow/cadence/issues/1171. 

@bjartek Great point that this is currently not possible, but is unrelated to what this FLIP is proposing.

turbolent:
>    - make sure all contracts follow the same mechanism when stating imports (currently that is not true even within onflow repo)
>    - make sure the contracts published that have imports have a way to communicate what that import is actually resolving to on different networks
>    - allow developers switch tools without having to replace import syntaxes and how those are being then resolved

You really nicely laid out the goals of this proposal, and I think it succeeds in addressing these goals, so +1 from me.
I don't think that this proposal needs to recommend or require any particular namespacing scheme.

joshuahannan:
Yeah, that makes sense. I'm fine with this then. 👍

PR #905, opened 2022-04-11
re-generate protobuf stubs

zhangchiqing:
## Description

This PR re-generate the protobuf stubs with the latest proto-gen-go (v1.27.1).

The reason to re-generate is that I need to update one of protobuf type definition (for block header), but re-generating the stubs for that changed block header type resulted lots of changes to other unchanged types, which isn't clear to be reviewed. So I separate it into two PRs. And this is the first PR to re-generate stubs with the latest tool. And the second PR will be updating the protobuf type definition for the block header as well as its stubs.
______

For contributor use:


jordanschalm:
Where are you getting the new version `protoc-gen-go@v1.27.1` from? I can only see a [version `v1.5.2`](https://github.com/golang/protobuf/tags) on the module we [link to in `flow-go`](https://github.com/onflow/flow-go/blob/e3a44f853679fd3895221652b7d980ebe08bafb7/Makefile#L71) and a [version `v1.2.0`](https://pkg.go.dev/google.golang.org/grpc/cmd/protoc-gen-go-grpc) in the [module we link to](https://github.com/onflow/flow/blob/abf014c8838ee14d31f6c69523fd00d60046b330/protobuf/README.md#L14) in the `README`.

Maybe we should update those links if the tool has moved.

PR #912, opened 2022-04-13
Add GetTransactionResultsByBlockID for EN API

m4ksio:


peterargue (approved):

smnzhu (commented):
Can we include the Access API changes as well so we only need to regenerate the protobuf once?

m4ksio:
> Can we include the Access API changes as well so we only need to regenerate the protobuf once?

Sure. Once its merged to master I can update this PR with fresh master

koko1123 (approved):

m4ksio:
Thanks @koko1123 for regenerating!

PR #934, opened 2022-05-03
FLIP: Interaction Templates

JeffreyDoyle:
# FLIP: Interaction Templates

## Abstract

This FLIP is a proposal for how:
- Templates of Flow transactions and scripts should be constructed.
- Transaction and script templates should be audited.
- Transaction and script template audits should be verified.
- Applications might use a Flow transaction or script template to carry out their underlying functionality

### DIFF
--- a/flips/20220503-interaction-templates.md
+++ b/flips/20220503-interaction-templates.md
@@ -1,0 +1,431 @@
+# Interaction Templates
+
+| Status        | Proposed      |
+:------------ | :----------------------------------------------------  |
+| **Status**  | Draft |
+| **FLIP #**         | [NNN](https://github.com/onflow/flow/pull/NNN) |
+| **Author(s)**  | Jeffrey Doyle (jeffrey.doyle@dapperlabs.com) |
+| **Sponsor**     | Jeffrey Doyle (jeffrey.doyle@dapperlabs.com)            |
+| **Updated**     | 2022-05-03                                         |
+
+## Abstract
+
+This FLIP is a proposal for how:
+- Templates of Flow transactions and scripts should be constructed.
+- Transaction and script templates should be audited.
+- Transaction and script template audits should be verified.
+- Applications might use a Flow transaction or script template to carry out their underlying functionality.
+
+## Background
+
+### Contract Interactions
+A common practice in the development of Flow contracts is to, along with the contracts themselves, also provide a prescribed suite of ways to interact with the contract(s) to accomplish various functionality. Developers who build software that interacts with these contracts can use these interactions in their projects.
+
+For example, alongside Flow's core staking contracts, [a suite of transactions and scripts are provided](https://github.com/onflow/flow-core-contracts/tree/master/transactions/stakingCollection) for developers who wish to interact with Flow's staking ecosystem. Apps like [Flow Port](https://port.onflow.org), block explorers, wallets, among others use these transactions and scripts to perform staking actions, and query state about Flow's staking ecosystem.
+
+This FLIP proposes that a higher order term for transactions and scripts be called _interactions_. Interaction intends to be a catch-all-term for something that "interacts" with the Flow blockchain to accomplish a goal. This term will be used throughout the remainder of the FLIP.
+
+### Cadence Interpretation
+Cadence has made incredible progress in furthering safe, secure, clear and approachable smart contract development on Flow.
+
+However, end users, developers and wallets alike can sometimes have difficulty interpreting what a cadence transaction or script might do when executed. For example, if an end user is presented with a malicious transaction to sign using their wallet, they likely will not have the technical ability to read the cadence transaction to determine that it is malicious.
+
+Wallet developers who wish to safeguard their users from malicious transactions can attempt to inference what a transaction might do before prompting it to be signed by a user. The problem with this approach is that predicting all the outcomes of a transaction is difficult, and an implementation today of such a prediction mechanic for all transactions would likely not be robust enough to provide sufficient user protection.
+
+A pattern that has emerged in the Flow community is for contract developers to have their transactions added to various repositories of known safe transactions. The maintainers of these repositories can audit these transactions for safety, and add them to repository if they are determined to be secure. An example of this in practice is the [Portto (Blocto) Flow-Transactions repository](https://github.com/portto/flow-transactions/blob/main/transactions) . When Blocto is asked to sign a transaction, it checks to see if the transaction exists in their Flow-Transactions repository. If it does, it then has greater confidence in that transactions safety, and makes that confidence known to their user though their UI.
+
+### Interaction Metadata
+A cadence script or transaction alone does not contain metadata about itself. Often, application and wallet developers may display a _human readable_ title or description about a transaction before requesting it to be signed by a user. They may wish to display a title and description about each argument it needs to collect from the user, or information about how the transaction will impact the user's account and assets. The metadata a developer might consume in their application could take many forms and be used for many purposes.
+
+An application or wallet may also want to know what a transaction will _do_ to protect against performing actions that will produce an undesirable result. For example, an application / wallet may wish to prevent a user from executing a transaction that sends more FLOW tokens from their account than they have, because that transaction will fail. An application / wallet may also want to what version of an interaction's dependency tree it was created against.
+
+Currently, applications and wallets are left to come up with their own metadata around the transactions they support. For example, on Flow Port, for each transaction it supports, it has included a title and description for it, along with titles and descriptions of each argument. Contract developers do not yet have a standardized way to provide metadata around their interactions that can be consumable by the applications and wallets that engage with them.
+
+## Objective
+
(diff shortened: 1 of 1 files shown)


pgebheim:
Thanks! Looking forward to reading through the proposal :-)

bjartek:
I read through the proposal and I really like the concept alot, however there is one part of it that I would advice a slightly different approach on. Right now the code that is stored contains variables that have to be replaced for each network, and then you store the addresses for the different networks. The problem with this is that when wallets receive the transaction these replacements are already made, so we make the work for wallets very hard. 

In stead I propose that we replace all addresses and just duplicate the code and hash for each network. 

I have done some work on this in overflow. 
https://github.com/bjartek/overflow/blob/main/overflow/parse.go you can call it like this

```

package main

import (
	"encoding/json"
	"fmt"

	"github.com/bjartek/overflow/overflow"
)

func main() {
	g := overflow.NewOverflow().ExistingEmulator().Start()

	res, err := g.ParseAllWithConfig(true, []string{
		"^admin*",
		"^setup_*",
	}, []string{})
	if err != nil {
		panic(err)
	}

	merged := res.MergeSpecAndCode()

	file, _ := json.MarshalIndent(merged, "", "   ")
	fmt.Println(string(file))
}
```

it will produce a json file like this
https://gist.github.com/bjartek/6a707eef7bb808b0325b95dd3bc919e9

This is an example for bl0x. 

We then publish an NPM module for that file that the frontend code uses.

bluesign:
I strongly feel this should be stored on chain.

- It would be more secure
- It would be more discoverable
- It would be more decentralized

Also I think some common tasks should be standard, Initialize new account etc

Ofc best would be to add new Storage to Flow called interactions. Second best would be a Registry contract to manage interactions on chain.

Making this onchain would also allow finally to use Ledger with FCL

Also I feel like with MetadataViews, if we store interactions on chain, we can reference interactions on resources. Send to another address this NFT , Add to collection etc

pgebheim:
Reminder: discussion about the content of the flip should be directed to the Forum linked in the FLIP.

This is to me preserve a single stream of communication across all in PRs and changes to a single flip.

https://forum.onflow.org/t/flip-934-interaction-templates/3080

bluesign:
@pgebheim I don't think forum is better for communicating FLIPS, I am not sure if here is the best, but for me it is 10 times better than forum. 

- Here I can see if FLIP is updated, what is changed etc. 
- Some FLIP owners are kind to communicate back 
- I can link from here code etc easier
- Changes are directly notified via GitHub notifications 

I don't know maybe active forum users ( if they exist ) disagree with me, I think forum is a bit dead town.

pgebheim:
A FLIP will have a multiple PRs as it moves through the process.

We want to keep all conversation about the content of the flips on one place and not spread across multiple threads in GitHub.

turbolent:
Nice work writing up a solution proposal for this problem Jeffrey! 
There is definitely a big need to improve the UX and such a standard would help with it.

I like the proposal's feature set, it is great that it includes internationalization.

My biggest concern / question is with the storage / provision of it:


> Contract developers should make available their Interaction Templates to others who wish to use them. 
> They can do this through providing them using a webserver, and have consumers query them. 
> They may also choose to store them on-chain. They may even choose to store them on IPFS. 

Do we want to encourage a non-chain solution here, so this becomes authoritative?
When a template is provided on-chain, and provided by the contract itself, it can be easily trusted.
When a template is provided off-chain, anyone can publish templates for anything, how can they be trusted / verified?


As for the concrete format of the template:

- Maybe improve the naming. Instead of shortening fields, just spell them out, e.g. `vsn` -> `version`
- What is the purpose of the `f_` prefix?
- What is the difference between the outer version and the inner "data" version?
- What is the "data"? What is the reason for nesting it – could it just be at the top-level?
- `data.arguments`: 
  - Could this just be an array, instead of an object where each entry has to specify an index?
  - Is the idea to allow lookups based on name?
- Keys in `messages` ... `i18n`, e.g `en-US`, should be strings if they are not identifiers, e.g. `"en-US"`

bjartek:
I am 100% behind an onChain solution for this. We want the trustless properties of a blockChain, we want the traceability it offers.

tarakby:
Just a quick comment that is unrelated to the core idea :) 
I noticed SHA-256 is used for IDs which is totally fine, except that all IDs in Flow structures (transaction IDs for instance) are computed using SHA3-256, so I thought using SHA3 may keep things more consistent within the ecosystem.

SaveTheRbtz:
a tiny nit picking: 

You may consider using URLs for locations in the doc, e.g:
```diff
- GET audits.trusted-auditor.com/id/{template_id} -> InteractionTemplateAudit
+ GET https://audits.trusted-auditor.com/id/{template_id} -> InteractionTemplateAudit
```
```diff
- DNS:transfer-flow.interactions.onflow.org
+ dns://transfer-flow.interactions.onflow.org
```
```diff
- IPFS:64EC88CA00B268E5BA1A35678A1B5316D212F4F366B2477232534A8AECA37F3C
+ ipfs://bafybeigdyrzt5sfp7udm7hu76uh7y26nf3efuylqabf3oclgtqy55fbzdi
+ ipfs://QmbWqxBEKC3P8tqsKc98xmWNzrzDtRLMiMPL8wBuTGsMnR
```
(not sure if template's location should be a string or a list of string to provide fallback on the client to have an ability to download template from multiple places.  Depends on whether the additional complexity worth the reliability benefit.)

Also, IIUC the purpose of encapsulating FLIP's content into the `data` object is to simplify the computation of the content hash (`id`)?  Maybe it is possible to archive the same by instead moving the content to the upper level (as @turbolent suggested) but incapsulate the metadata into the `.metadata` object?

flips/20220503-interaction-templates.md
```diff
@@ -0,0 +1,446 @@
...
+                    testnet: {
+                        address: "0x9a0766d93b6608b7",
+                        fq_address: "A.0x9a0766d93b6608b7.FungibleToken",
+                        pin: "asdfasdfasdfasdfasdfasdfsadf123123123123",
+                        pin_block_height: 10123123123
+                    }
+                }
+            }
+        },
+        arguments: {
```
bthaile, line 187: Should arguments be typed? example: amount be `ufix64`. I wouldn't expect the type be parsed from cadence script.

flips/20220503-interaction-templates.md
```diff
@@ -0,0 +1,446 @@
...
+                    },
+                    description: {
+                        i18n: { // Internationalised (BCP-47) set of human readable messages about the argument
+                            en-US: "Amount of FLOW token to transfer",
+                            fr-FR: "Quantité de token FLOW à transférer",
+                            zh-CN: "要转移的 FLOW 代币数量"
+                        }
+                    }
+                },
+                balance: "0xFUNGIBLETOKENADDRESS.FungibleToken" // The token this argument acts upon.
```
bthaile, line 206: `balance` seems to not fit here, would all arguments have "balance" property?

flips/20220503-interaction-templates.md
```diff
@@ -0,0 +1,446 @@
...
+
+Contract developers may also choose to store their Interaction Templates behind static identifiers. This allows for the implementation of a Interaction Template to change, while the identifier stays the same. For example, a contract developer may choose to host a webserver and serve an Interaction Template behind a static url:
+
+```
+GET interactions.nft-project.com/purchase-nft -> InteractionTemplate
+```
+
+Consumers of this Interaction Template could query the static identifier, and always get the most up to date Interaction Template stored there.
+
+### Interaction Audits
```
bthaile, line 298: Based on a forum conversations, there is a strong use case to have template audit information on-chain. I think it can be removed from the actual template.

flips/20220503-interaction-templates.md
```diff
@@ -0,0 +1,446 @@
...
+
+At a high level, Interaction Templates attempt to standardize how contract developers, wallets, users, auditors and applications, can create, audit and verify the intent, security and metadata of Flow scripts and transactions.
+
+### Metadata Standardization
+By standardizing **Interaction Template Metadata**, the Flow community can start to build applications that can consume metadata, instead of having to come up with it themselves for each interaction they need to support. Since the format of the metadata would be consistent, this enables parties to support a much wider number of interactions, including any number of arbitrary interactions.
+
+Standardizing metadata helps all parties involved in the execution of an interaction to better understand what the interaction requires and does. For applications, metadata allows them to understand and present the interaction to a user in an intuitive way. For wallets, metadata also helps them understand what a transaction will do to an account, and gives them paths to prevent undesirable outcomes and reject malicious transactions. For users, metadata helps to promote human readable understanding of the impacts of a transaction.
+
```
bthaile, line 48: In your examples, I'd suggest adding optional Metadata that the developer can add to convey information to Auditor and/or Wallet. This would better show the flexibility of the standard, it'll be very difficult to handle all data cases in this flip.

flips/20220503-interaction-templates.md
JeffreyDoyle: No not all arguments would have a "balance" property. What this aims to do is create a link between the dependencies of the interaction and the argument to show they're related. With this property an app/wallet might be able to use it to determine if a user has sufficient balance of a FT to successfully execute a transaction, for example.

pgebheim:
Merging as draft

pgebheim (approved):
👍 Formatting good for draft!

PR #941, opened 2022-05-16
Add FLIP to propose separating reference expressions into two distinct operations

dsainati1:
This FLIP proposes to address a common source of confusion among Cadence users by decoupling the two parts of the reference expression: creating the reference and casting it to a type. Under this model, the `as` operator would only be a casting operator, while references would be created with `&v`. 

Related work to this is https://github.com/onflow/cadence/pull/1566, which bans the use of `as!` and `as?` in reference expressions in the parser, as these do not function at runtime. 

______

For contributor use:


flips/20220516-reference-creation-semantics.md
```diff
@@ -0,0 +1,269 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [941](https://github.com/onflow/flow/pull/941)|
```

dsainati1:
Thanks for writing this up! I agree with your thoughts on `auth` references; while the ability to downcast certain references is useful, I agree that needing to manually specify this with the `auth` keyword feels weird somehow, and should ideally be automatically determined by the checker. Changes to `auth` references would be beyond this scope of this particular FLIP however, but we can return to this topic in the (hopefully near) future. One quick note though, the checker does indeed check auth/non-auth references; this code passes typecheck at runtime:
```
let x = 3
let y = &x as auth &AnyStruct
let z = y as! &Int
```
While this code does not:
```
let x = 3
let y = &x as &AnyStruct
let z = y as! &Int
```

As for your suggestion for restrictions being used for references, is that meant to be in addition to or as a replacement for the requirement to provide explicit types for resource references.

dsainati1:
Hmm, I think the syntax you propose is confusing for a different reason; normally the restricted type syntax `T{X, Y}` here is only used on types; but your example has it being used on a value `MySuperCollection` to get a reference to it. This I think is conflating values and types by allowing you to use a type-specific syntax on a value.

turbolent:
Thank you for this proposal Daniel! I like it a lot.

I definitely agree with the motivation. The current syntax of reference expression trips up many developers who are learning Cadence, and even developers who have been writing Cadence code for a while.

Before commenting on the proposal, I wanted to quickly add some context and design rationale.

When we originally designed the referencing feature, we made the following observation:

Unlike in other languages, references in Cadence are mostly used for sharing access to untrusted code. 
This is done in tandem with restricted types.

For example, a developer might want to share access to their FT vault resource, typed `@FungibleToken.Vault`, but restrict access to just the functionality exposed through the `FungibleToken.Balance` interface,  by creating a reference `&FungibleToken.Vault{FungibleToken.Balance}`.

As the reference is shared with untrusted code, that code should not have the ability to increase the level of access to the shared resource, e.g. by casting to a less-restricted type like `&FungibleToken.Vault`.

This is why reference types are non-authorized "by default", and the more powerful variant, authorized (for lack of a better word) references, requires a keyword.

Authorized references are basically references like in other languages, where the owner of the reference has arbitrary access to the target. This is uncommon in smart contracts, and actually a potential for a foot gun, though there are cases where this is useful.

For example, if a developer owns certain objects, for example NFTs in account storage, they might want to access them an perform certain type-specific operations on them:
An account's storage might store an array typed `[NonFungibleToken.NFT]` – by borrowing using an authorized reference, they can downcast to the concrete NFT type and call a function unique to that type.

In addition, we were worried that creating a reference with a too powerful type would be too easy. This is why the type must always be provided in a reference expression, even if it is the same as the type of the value. It is verbose, what makes the developer be explicit and think about what kind of reference they are creating.

For example, a developer might want to share access to their FT vault resource `vault`, typed `@FungibleToken.Vault`. If it were possible to create a reference with just `&vault`, the developer might accidentally share full access to the reference in cases where they might not intent to.

So far these two design decisions have helped avoid many security problems:  Authorized references are rightfully uncommon, and we have not heard of cases where untrusted code was accidentally given more access than necessary leading to an exploit. I would argue that it worth persevering these two design decisions to some degree.

Now to this proposal in particular:

I am not quite sure if the proposal suggests making a difference between taking a reference to a resource-minded value or a struct-kinded. If it does, I would suggest not special casing any kind and making reference expression work the same for all objects, regardless of the kind. (Maybe this is not suggested and I misunderstood)    

I really like the idea of making reference expressions behave similar to empty container literals (empty array `[]` and empty dictionary `{}`). For such literals, the type checker infers *no* type and requires the developer to provide an explicit type annotation. For example, for `let array = []` the type checker requires an explicit type annotation, 
which can be provided by the developer e.g. through a static type annotation (`let array = [] as [Int]`), 
or e.g. through a type annotation of a variable declaration (`let array: [Int] = []`).

Making reference expressions behave this way removes a special case from language, and allows developers to translate the intuition they gained from type annotations and static type annotations to any expression, including references.

Regarding the need for an `auth &` operator in addition to the `&` operator: This should actually not be necessary. By changing the type checker not to infer a type for the reference expression, a type must be provided. That type may be an authorized type. For example, `let ref: auth &S = &s` or `let ref = &s as auth &S` will just work.

The nice side-effect is that this keeps all current code working as before.

I was not quite sure if this could be implemented and how it would affect code, so I opened this draft PR which implements this proposal, and also shows how no new `auth &` operator is needed: 
https://github.com/onflow/cadence/pull/1661

I also want to add that I prefer the proposal's solution over the alternatives that were considered before.

robert-e-davidson3 (approved):

bluesign:
> Hmm, I think the syntax you propose is confusing for a different reason; normally the restricted type syntax `T{X, Y}` here is only used on types; but your example has it being used on a value `MySuperCollection` to get a reference to it. This I think is conflating values and types by allowing you to use a type-specific syntax on a value.

You are right this is confusing in current situation. But I think it is because we defined things like this. 

But isn't the point of getting reference is like this?

I mean if C is composite of type T:

` &C is type of:  &T`.  ( here we accept that first &  is operator second one is type specifier ) This is not disturbing our eyes, but actually it has the same problem you pointed out. 

so for me : 

` &C{X, Y} is type of: &T{X, Y}` doesn't feel that wrong. Actually feels more natural. 

But to be honest, I don't have strong feeling on this, after reading @turbolent's comment about [], I am in favor of proposed too. 

But my point about somehow automatic auth stands, though it is not in the context of this FLIP. 

And again thanks for fixing this

dsainati1:
> The nice side-effect is that this keeps all current code working as before.

@turbolent In theory this means we could release this now right? If it's not a breaking change we wouldn't have to wait until Secure Cadence.

turbolent:
@bluesign I was also confused about it before (mainly because of the casing), but I think I get now what you are proposing, a short-hand syntax for taking a reference with a restricted type? This is a good idea, though maybe it should be a separate FLIP which could be added in addition to this one?

turbolent:
@dsainati1 I looked into this a bit more: 
The caveat is that existing *correct* code keeps working, i.e. which uses the correct syntax `&v as &T`. 

Code which is currently using the invalid (but unfortunately accepted/ignored) operators `as?` and `as!`, e.g. `&v as! &T` or `&v as? &T`, will get rejected, as the reference itself is missing an explicit type (the casting expression does not suggest a type to the casted expression). 

However, we were planning to fix this anyway, so I don't feel like it is a blocker.

SupunS:
I've always been a fan of having just `&v` for a reference expression, so I'm totally in favour of the proposal.

The only concern I had was the `auth` references - didn't have any clean solutions/suggestions for that. But I like the idea of getting the 'authorize'-ness from the context (i.e: cast or the contextually expected type), and if the context doesn't have it, then being explicit. It nicely stands with the other semantics of the language.

flips/20220516-reference-creation-semantics.md
```diff
@@ -0,0 +1,191 @@
...
+let p = &r // type error: `&r` does not have a type provided
+```
+
+Authorized references can be created the same way by using the `auth` keyword as normal.
+
+```cadence
+resource R { /* ... */ }
+
+let r <- create R() // r has type @R
+let p = &r as auth &R // p has type auth &R
```
SupunS, line 117: Maybe also add an example for auth-reference where the type is coming from the context.
e.g:
```cadence
let p: auth &R = &r
```

flips/20220516-reference-creation-semantics.md
```diff
@@ -0,0 +1,191 @@
...
+syntax when declaring functions. 
+
+### C/C++
+
+In C and C++, `&` is the referencing operator, while `*` is the dereferencing operator. The
+means that `&v` will create a reference to the value `v`, while `*p` will obtain the value pointed-to
+by the pointer `p`. Confusingly, C uses the `*` postfix to denote a pointer type, so if `v` has an `int`
+type, then `&v` has type `int*`.
+
+Other languagues derived from C like Rust have also chosen this syntax, although in Rust's case for a
```
SupunS, line 53: Another example is Go (similar to Rust)

SupunS (approved):

bluesign:
> @bluesign I was also confused about it before (mainly because of the casing), but I think I get now what you are proposing, a short-hand syntax for taking a reference with a restricted type? 

@turbolent Yeah it can be very useful. 

But my main target was how to prevent accidental resource references ( like in the example of the FLIP ) Currently we are requiring casting. But I feel casting and referencing should be distinct operations. ( Though your comment on empty array convinced me a bit ) 

So I thought what if referencing requires explicit restriction. then you have to somehow declare explicitly if you want reference without any restrictions. ( as `auth` keyword `WithoutRestrictions` keyword will not be much used ). Something like:  `&MySuperCollection{WithoutRestrictions}` 

Mostly I will use something like this:
`var receiverRef = &MySuperCollection{FungibleToken.Receiver} ` 

Benefit will be referencing will be totally free of casting and I don't need to declare type. 

Also now I can separate referencing and casting. 

```
var receiverRef = &MySuperCollection{FungibleToken.Receiver} 
var genericReceiver  = receiverRef as &AnyResource{FungibleToken.Receiver} 
```

turbolent (approved):

joshuahannan (approved):
Seems pretty straightforward to me! I support this for sure. 👍

dsainati1:
The implementation for this is approved (https://github.com/onflow/cadence/pull/1661), should we approve and merge this FLIP as well?

turbolent:
The implementation PR is only provided as a means to evaluate the proposal, so should not influence the decision if the proposal is accepted or not.

As for the approval of the proposal itself, we will discuss it in a meeting on Thursday. As always, feedback is still very welcome

turbolent:
@dsainati1 as discussed, given there has not been any further discussion for a week, we can probably consider this accepted

flips/20220516-reference-creation-semantics.md
```diff
@@ -0,0 +1,200 @@
+# Title of FLIP
+
+| Status        | (Proposed)       |
```
dsainati1, line 3:
```suggestion
| Status        | (Accepted)       |
```

dsainati1:
Nice! @turbolent shall we merge your implementation into Cadence now?

PR #798, opened 2022-02-03
FLIP: Capability Controllers

janezpodhostnik:
## Description

Cadence encourages a [capability-based security](https://en.wikipedia.org/wiki/Capability-based_security) model as described on the Flow [doc site](https://docs.onflow.org/cadence/language/capability-based-access-control). Capabilities are themselves a new concept that most Cadence programmers need to understand, but the API for syntax around Capabilities (especially the notion of “links” and “linking”), and the associated concepts of the public and private storage domains, lead to Capabilities being make this concept even more confusing and awkward to use. This proposal suggests that we could get rid of links entirely, and replace them with Capability Controllers (henceforth referred to as CapCons) which could make Capabilities easier to understand, and easier to work with.

PR #945, opened 2022-05-18
Flip for identity on auth account

bjartek:


bluesign:
> One of the most asked questions in the discord server is "how do i know what user called this method". 

Instead of putting `&Identity` to function signatures, why we don't give access to information with some Environment Information function like `getTransaction` and have a field like authorizers ?

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+| **Sponsor**   | AN Expert (core-contributor@example.org)      |
+| **Updated**   | 2022-05-18                                    |
+
+## Objective
+
+One of the most asked questions in the discord server is "how do i know what user called this method". In solidty the answer is "msg.sender", that pattern does not work on flow and it is not recommended to do things that way. 
+
+One of the reasons it is not recommended is that there is no secure way of doing that. 
+
+This FLIP aims to create an way to do it securely. Note that the FLIP does not cover when you should use this instead of other patterns for identifying the sender, it only covers that _if_ you need to do it you have a save way to do si.
```
pgebheim, line 16:
```suggestion
One of the most asked questions in the discord server is "how do i know what user called this method". In solidty the answer is "msg.sender", that pattern does not work on flow and it is not recommended to do things that way. 

One of the reasons it is not recommended is that there is no secure way of doing that.  This FLIP aims to create a way to do it securely.

*Note: This FLIP does not cover __when__ you should use this instead of other patterns for identifying the sender, it only provides a way to do it safely  _if_ you need to do it.
```

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Motivation
+
+Make is secure to identify that a particular auth account signed a transaction
```
pgebheim, line 20:
```suggestion
Add a secure to identify if  a particular auth account signed a transaction. 
```

I'd add some real motivations here, e.g. what are the cases where someone really *does* need to do it this way?

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+
+## User Benefit
+
+It makes it possible to use patterns that are common to solidity developers in flow
```
pgebheim, line 24:
```suggestion
It makes it possible to use patterns that are common to Solidity developers in Flow.
```

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+
+## Design Proposal
+
+The design proposal and a implementation is provided in the following PR
+https://github.com/onflow/cadence/pull/1363/files
+
+### Drawbacks
+
+It makes it possible to do something that it is documented is not best practice. Should this documentation be changed?
```
pgebheim, line 33:
```suggestion
Currently the documentation states that this pattern is not a best practice. Should this documentation be changed?
```

I still feel like we need more argument as to why this shouldn't be an anti-pattern and should get first class support.

pgebheim (changes requested):

flips/20220518-authaccount-identity.md
turbolent, line 16: @dete  brought up a good example use-case for where this is useful: solving the bootstrapping problem of sharing a capability. It might be nice to add it as an example. 

In general, it might be good to point out that capability-based access control should be preferred.

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+A user will be able to call `authAccount.identity` and get something that can verify that account but cannot be obtained any other way
+
+
+## Related Issues
+
+https://github.com/onflow/cadence/pull/1363/files
+
+## Prior Art
+
+Solidity user msg.sender to identify who as signed a transaction
```
turbolent, line 55:
```suggestion
Solidity provides the `msg.sender` field which allows identifying who signed the currently running transaction.
```

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+
+It makes it possible to do something that it is documented is not best practice. Should this documentation be changed?
+
+### Tutorials and Examples
+
+See PR for examples and changed docs
+
+### User Impact
+
+A user will be able to call `authAccount.identity` and get something that can verify that account but cannot be obtained any other way
```
turbolent, line 46:
```suggestion
A user will be able to call `authAccount.identity` and get a token representing the ability to access to an account, without it being obtainable any other way, i.e. it cannot be forged, as there is no constructor function for the  `AuthAccount.Identity` type.
```

turbolent (approved):
I think this would be a helpful addition to the language. 

We should emphasize that this feature should not be used for typical access control use-cases, and capability-based access control should be preferred. 

The proposal itself needs some editing, @bjartek could you please do a second pass incorporating the feedback? Thanks!

bluesign:
> We should emphasize that this feature should not be used for typical access control use-cases, and capability-based access control should be preferred.

To be totally honest, when you have this, using capability based access seems little stupid. This is too powerful. As I said before when we put resources owner, game was already over. 

There is no need to still try to push capability based control. It is a lost cause.

flips/20220518-authaccount-identity.md
```diff
@@ -0,0 +1,55 @@
...
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #) |
+| **Author(s)** | Bjate Stien Karlsen (bjarte@find.xyz)         |
+| **Sponsor**   | AN Expert (core-contributor@example.org)      |
+| **Updated**   | 2022-05-18                                    |
+
+## Objective
+
+One of the most asked questions in the discord server is "how do i know what user called this method". In solidty the answer is "msg.sender", that pattern does not work on flow and it is not recommended to do things that way. 
+
+One of the reasons it is not recommended is that there is no secure way of doing that.
```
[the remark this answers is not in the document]
bjartek: So should we incorporate this into the text?

flips/20220518-authaccount-identity.md
bjartek: Can you elaborate more on @dete´s example. I am not sure i get it. 

But after that yes I can add abit more meat on why cap based is preferred.

bjartek:
I have commited all of the suggestions will see if I can find time next week to do a more in-debt pass over this. 

@bluesign I still think there are rooms for both options TBH. This is kind of like a last resort.

bluesign:
@bjartek i think this is really good btw. I meant there shouldn’t be push for capability when it is not needed. A lot of contract code can be simplified with this. ( with security in place )

Though I also think scoping this can be little tricky. It should be valid on the first call scope only. ( contract function shouldn’t be able to pass this to another function of another contract )

flips/20220518-authaccount-identity.md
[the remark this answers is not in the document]
turbolent: Yes, please do

flips/20220518-authaccount-identity.md
turbolent: @dete Could you please provide more details?

dete: Here is the problematic use case. I'll describe possible solutions (including one using an Identity object) afterwards.

---

I have a project that requires some amount of human oversight, maybe something as simple as a "pause" button that lets me block activity in the smart contract temporarily in response to a bug or market panic. As such, I have an administrator capability in my account that gives me access to call the "pause" function if necessary.

In time, Bjarte joins my team, and I decide that it'd be useful if his account could also call the "pause" function in case I'm sleeping when the emergency happens. I'd like to send him a Capability for the same Admin object that I have. How do we get the Capability into his account securely?

---

One possible solution: Co-signing

One solution (which is possible in Cadence today) is for Bjarte and I to both sign a single transaction. That transaction can create/copy the Capability using my account, and then store it in his account.

This solution is workable, but awkward. In particular, it has a coordination problem. Both Bjarte and I have to be online at more-or-less that same time in order to both sign the shared transaction within the expiry window.

---

Another solution: The Dropbox

Another answer would be for me to create the Capability for Bjarte whenever it suits me and then "leave it somewhere he can get it at his convenience". But where do I leave it that doesn't allow a stranger to intervene?

Ideally, I'd have a public function on my account that returns that Capability, but only if the caller provides Bjarte's Identity.

---

To be clear: I don't love this use case, because once the Identity object exists people might start using it in other cases, and if it's used frequently, in a variety of ways, it could open the window for attacks. For example, perhaps a malicious actor notices that I've posted a Capability for Bjarte before Bjarte himself notices it. That attacker could offer him "1000 free FLOW!" using a different Dropbox object that asks for his Identity. But instead of returning the 1000F, the malicious Dropbox object uses Bjarte's Identity to grab the Admin Capability I was offering and transfer it to the attacker!

The problem with an Identity object in a capability-based security system like Cadence that sharing the Identity object is _transitive_. I'm not just proving my identity to a smart contract when I pass it my Identity, _I'm temporarily allowing it to impersonate me_!

We'd really need wallets to carefully restrict the use of the Identity object to scenarios where they explicitly check the type of the Resource they are passing the Identity to _in the transaction itself_, and ensuring that the type they see is on a short list of known-safe Dropbox implementations (ideally one!).

Unfortunately, I don't really have a better solution for bootstrapping a new Capability setup that doesn't suffer from the coordination problem of double-signing. I'd love to hear other suggestions!

turbolent:
After further discussion it is the consensus that this proposal should be rejected. 

The language feature was proposed mostly as a solution to the bootstrapping problem of sharing a capability. The feature would indeed solve this problem, but would likely also be used to implement identity based access control.

Instead, the plan is to create a new FLIP for a solution to the capability bootstrapping problem. This ties in nicely with the ongoing effort to improve the capability API in FLIP #798. 

Thank you everyone for your feedback, a productive discussion, and especially @bjartek for the prototype implementation.

PR #1030, opened 2022-07-04
FLIP-934 Interaction Template Updates

JeffreyDoyle:
## FLIP-934 Interaction Template Updates

Changes

- Removes `version` of template
- Adds serialization spec for `InteractionTemplate`, `InteractionTemplateInterface` and `InteractionTemplateAudit`
- Adds additional diagrams explaining templates workflow


bluesign:
Hey @JeffreyDoyle , great update!

btw on the issue of serialization, and comment from lilico wallet recentlyy on forum made me thinking and one idea came up in my head. 

What if we make subsections hashable ( instead of full structure ) and allow dapp to send proof with partial content instead. So we can make FCL and other wallets to support without grabbing interaction templates from somewhere else. 

template-message
template-dependency
template-cadence
template-arguments

etc can all have hashes of the content. Then we can find a root hash from them. 

I am thinking more like : 

template-encoded              = RLP([ 
    sha3( template-type ), 
    sha3( template-interface ), 
    sha3( template-messages ), 
    ..
])

even we can separate messages maybe, so dapp can provide EN-US translation and combined hash of others to provide proof.

pgebheim (approved):

PR #1056, opened 2022-07-15
FLIP for purity/mutability analysis and restriction in function conditions

dsainati1:
Implementation of https://github.com/onflow/cadence/issues/1806 (further discussion can be found there). 

## Description

This FLIP proposes to add a new syntax and semantic analysis to determine which functions in Cadence have side effects, and which are `pure`. This analysis would allow us to then restrict or limit the contexts in which side effects can occur (see the "Motivation" section of the FLIP for a non-exhaustive list of issues and features where this would be useful).

A work-in-progress implementation of this FLIP for demonstration and proof-of-concept purposes can be found here: https://github.com/onflow/cadence/pull/1818

______

For contributor use:


flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,292 @@
+# Function Purity 
+
+| Status        | (Proposed)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [1056](https://github.com/onflow/flow/pull/1056) |
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,292 @@
...
+fields; nested composites could not be enforced because we did not have a mechanism for knowing what 
+composite methods were "mutating". However, if we could know whether a function or method is `pure`, we
+could restrict the calleable methods on such fields to be only `pure` ones. 
+
+Similarly, a [FLIP](https://github.com/onflow/flow/pull/1043) and an [issue](https://github.com/onflow/cadence/issues/1304)
+that suggest improvements to references to make them less powerful and prone to safety-footguns suggest to limit the 
+operations that can be performed on references, or to allow the creation of read-only references. However, these
+restrictions would only be enforceable at runtime, since we do not have a way statically to decide whether an operation is
+read-only. However, if we could know whether a function or method is `pure`, we could require that such references only 
+perform `pure` operations or are passed to `pure` functions.
```
SupunS, line 34: Would be nice to have some examples of how this FLIP would solve the other issues/FLIPs mentioned in this section

SupunS (commented):
This is a nice writeup!

I'm also curious on how would this affect the user experience. As in, how easy/hard would it be for a dev to write a program/contract that has mix uses of pure/impure functions? Perhaps,  can try migrating an existing contract (e.g: one of the core contracts, or maybe TopShot contract?) to make use of these annotations, just to get a sense on the UX.

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,292 @@
...
+    ```
+
+    while the following are not:
+
+    ```cadence
+    pure fun foo(a: [Int]): Int {
+        a.append(3)
+        return a.length
+    }
+    ```
```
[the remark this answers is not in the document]
dsainati1: Ah ok, so it seems like we should be able to allow function arguments to have mutations since they are copied. However, the core of this example remains the same if `a` is captured from outside; so I will change the FLIP to demonstrate this instead.

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,292 @@
...
+
+    Note that it is not always possible to determine statically whether or not a mutation is occuring on a locally or externally scoped variable, 
+as in this case, for example:
+
+    ```cadence
+    pure fun foo(a: [Int], i: Int): Int {
+        let b: [Int] = []
+        let c = [a, b]
+        c[i].append(4) // it is impossible to tell statically whether it is `a` or `b` that is receiving the write here
```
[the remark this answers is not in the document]
dsainati1: Same as above.

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,292 @@
...
+Function types can also have purity annotations, to be placed after the opening parenthesis but before the parameter list. So, for example, these are valid
+types:
+
+```cadence
+    let f: (pure (Int): Int) = ...
+    let g: (impure (String): String) = ...
+    let h: (pure (): (impure (): Void)) = ...
+```
+
+Any function types without an explicit purity annotation will be considered `impure`.
```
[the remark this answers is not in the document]
dsainati1: Purity inference would only infer purity for unannotated function values. Function types (like `(():void)`) won't be able to have their purity inferred this way, so they need a sensible default.

bjartek:
I like this FLIP!

bluesign:
I second @SupunS 's proposal, I have really big doubts about the gain for developer here. 

As an average developer what I will gain from this ?

Cannot this be implemented invisible to user with runtime error ? I don't see any mention of that

> All of the issues in the Motivation section do have alternative solutions; if we elected not to proceed with this proposal we could pursue those instead.

I think that section is for explaining those alternatives, not confirming of their existence. 


I suggest, if we can do purity analysis ( we should be able to as we cannot trust developer marking pure) , I prefer runtime error when purity is needed. It will be a rare case, and panic has strong feedback loop when developing. I think this over complicates life of the developer as-is.

dsainati1:
> As an average developer what I will gain from this ?

This will actually make it simpler, not harder, to think about your code. Reasoning about how and when mutation occurs in code is famously difficult because in most cases it requires knowing exactly how functions are implemented, even those that you didn't write. For a developer to able to look at a function declaration at a glance, see the `pure` keyword, and know with certainty that calling this function will not have any unexpected side effects they need to account for, will reduce the amount of mental overhead necessary to write composable code. 
 
> Cannot this be implemented invisible to user with runtime error ? I don't see any mention of that
> I suggest, if we can do purity analysis ( we should be able to as we cannot trust developer marking pure) , I prefer runtime error when purity is needed. It will be a rare case, and panic has strong feedback loop when developing. I think this over complicates life of the developer as-is.

I disagree pretty strongly that a runtime error is preferable to a static error in almost all cases. It's better for correctness, as a static error guarantee that a potential mistake is caught, while a runtime error can slip through testing if it only occurs with a specific set of inputs. It's also better for speed and ease of development: iterating on a static error is simple because the type checker will tell you exactly where it occurs and how to fix it, while a runtime error can require hours of debugging to even diagnose the source. A classic example of this tradeoff is in optional types (like Cadence and TypeScript have) where by adding an extra bit of complexity to the language in making `nil`-able values have a different type to normal values, we can prevent an entire category of runtime errors from even occurring and help developers more easily reason about where `nil` can occur in their code. 

There is also the consideration that adding new runtime errors to existing contracts is significantly harder to deal with for developers. Adding a new static error is simple, because developers can fix the code that is reported as broken and redeploy their code. If we instead add a new runtime error though, developers may not even know that their code's behavior may change in a specific set of circumstances, and only encounter a critical error weeks or months after the Flow upgrade. It is more developer friendly for contracts to break early so they can be fixed in advance of a spork than for contracts that seem to be correct to suddenly break in production. 

> > Swift takes the opposite approach, having a mutating keyword that functions must use in order to be allowed to mutate program state, essentially implying that functions are pure by default. Given the large amount of existing code in Cadence however, adoption will be easier by default if impurity is the default assumption.
> If we will somehow do this, I think Swift has the best approach for user friendliness, it is pretty easy to declare that 'you will modify something here' than declare 'you will `not` modify something here' .

I am not against making `pure` the default and requiring users to declare functions to be `mutating`; I think you may be right that the end result may be better for the language. My primary concern with making `pure` be the default, however, is that this will dramatically increase the amount of code that would be broken by this change, which could make it challenging for users to adopt.

bluesign:
> For a developer to able to look at a function declaration at a glance, see the pure keyword, and know with certainty that calling this function will not have any unexpected side effects they need to account for. 

Unless the owner of that contract changed that to impure ( which will break all my contract, as checker will fail ) Not broken when it is used. ( runtime error case ) 

I mean not only my contract, every collection containing a single of my NFT probably, every contract importing my contract. 

> Adding a new static error is simple, because developers can fix the code that is reported as broken and redeploy their code.

yeah but it will also break running contracts ( like Cadence last breaking change did with optional references ) 

Oh I mean maybe off-topic, but on optionals subject, please check transactions and contracts on mainnet. You will understand what I mean, proper optional usage is really really low. Panic is everywhere. Composable stuff is not possible at all.

dsainati1:
After some internal discussion, we identified one potential alternative approach to this that would be significantly more lightweight than the idea in the main FLIP. 

This version would only be designed to address the function pre/post condition use case, and would feature no inference. Instead, all functions would be impure by default, and only by annotating a function as `pure` would purity be possible. This would only be required for functions that are designed to be called from pre/post conditions, which is unlikely to be a large number of use cases, and also unlikely to result in long function dependency chains. 

This lightweight of a feature would probably not be suitable for any heavy weight use cases like mutability of fields or references, but could be a good solution specifically for the function pre/post condition use case.

SupunS:
I like this lightweight version more. The biggest concern I had with the original proposal was the UX/composability. But by limiting the purity requirement to only pre/post conditions, and by making it explicit (no inference), I think this would have a minimum (almost zero) impact on UX/composability.

> This would only be required for functions that are designed to be called from pre/post conditions, which is unlikely to be a large number of use cases, and also unlikely to result in long function dependency chains.

Fully agree with this.

In future, if we ever decide that purity is needed for other parts of the language (e.g: original proposal), then this would be a good foundation for that.

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+```cadence
+pub resource interface R {
+    priv pure fun foo(): Void {}
+    pure init()
+}
+```
+
+Any functions that do not have a `pure` annotations will be considered impure.
```
robert-e-davidson3, line 56:
```suggestion
Any function that does not have the `pure` annotation will be considered impure. 
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+```cadence
+    let f: (pure (Int): Int) = ...
+    let h: (pure (): (pure (): Void)) = ...
+```
+
+Any function types without an explicit purity annotation will be considered impure.
+
+### Type Checking
+
+A function that is `pure` (if it was explicitly declared with the `pure` keyword) may not perform any operations with side effects
```
robert-e-davidson3, line 70:
```suggestion
A function that is `pure` (it was explicitly declared with the `pure` keyword) may not perform any operations with side effects 
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+        let c: [Int] = &[] as &[Int]
+        if b {
+            c = &a as &[Int]
+        }
+        c.append(4) // it is impossible to tell statically whether `a` is receiving the write here
+        return a.length
+    }
+    ```
+
+Any function declared to be `pure` will enforce that none of these operations occur in their bodies. If such an operation does occur, then the checker will report
```
robert-e-davidson3, line 129:
```suggestion
Any functions declared to be `pure` will enforce that none of these operations occur in their bodies. If such an operation does occur, then the checker will report
```

robert-e-davidson3 (approved):

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+is the default assumption. 
+
+## Questions and Discussion Topics
+
+* Should `pure` functions be allowed to emit events and log data? These are technically side effects, but
+they are not problematic for the specific use cases we have identified for this analysis. Are there any other
+potential use cases for this analysis that would require events and logging to be restricted in `pure` contexts?
+
+* Given that `pure` in Cadence is closer to `view` in Solidity, a language with which our users may be familiar, 
+should we instead consider using `view` instead of `pure` in our implementation of this feature?
```
turbolent, line 211: 2 cents: I quite like Solidity's keywords/definitions, as they match more closely the typical definitions of what is considered a pure function, I'd go with `view`

turbolent (commented):
Nice! I quite like this new variant of the proposal

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+        return a.length
+    }
+    ```
+
+    Cases like this will have to be rejected statically for safety.
+
+* Writing to storage. This includes operations like `save` and `load` (since if a resource is loaded it will be moved out of storage), as well as `unlink`, but
+not operations like `copy`, `borrow`, `type`, or `getCapability`.
+
+* Writing to account state. This includes operations like adding, updating or removing a contract and adding, updating or removing account keys.
```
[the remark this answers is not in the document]
dsainati1: Yes; an alias would have the same type as the function it is aliasing, so any alias for an impure function will also be impure.

flips/20220715-cadence-purity-analysis.md
dsainati1: cc @SupunS @robert-e-davidson3 @bluesign @bjartek how do you all feel about this. I am ambivalent either way, but I want to get buy in from a majority of the FLIP commenters before i make the switch.

bluesign: `view` seems better for me personally.

turbolent:
I guess the "Breaking Change" label can be removed, as the latest version of the proposal is just an addition?

Or is changing conditions to only allow calls of pure functions part of this proposal? If so, it should be maybe made more explicit, or could be a follow-up proposal.

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+```cadence
+pub struct S {
+    pub view fun foo(): Void {}
+    view init()
+}
+```
+
+```cadence
+pub resource interface R {
+    priv view fun foo(): Void {}
```
turbolent, line 51: Just as a side note for the implementation: The type checker will need to ensure that the function in the concrete type that implements the interface has the modifier it is given for the function in the interface. This conformance has to be checked for interfaces and nested type requirements

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
+# Function Purity
```
turbolent, line 1: Given that the proposal is not about pure functions anymore, maybe rename this proposal / language feature to "View functions", [similar to Solidity](https://docs.soliditylang.org/en/v0.8.13/contracts.html#view-functions).

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+| **FLIP #**    | [1056](https://github.com/onflow/flow/pull/1056)     |
+| **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Updated**   | 2022-08-25                                           |
+
+## Objective
+
+This FLIP proposes to add a new syntax and accompanying semantic analysis to determine or 
+enforce that a given function or method is "view", that is, lacking in side effects. We would
+then add an additional checker rule to enforce purity in function conditions.
```
turbolent, line 14:
```suggestion
enforce that a given function or method is "view", that is, lacking in side effects.
 We would then add an additional checker rule to enforce that function conditions only call view functions.
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+pub resource interface R {
+    priv view fun foo(): Void {}
+    view init()
+}
+```
+
+Any function that does not have a `view` annotation will be considered impure. 
+
+Function types can also have purity annotations, to be placed after the opening parenthesis but before the parameter list. So, for example, these are valid
+types:
```
turbolent, line 59: Like for the name of the proposal/feature, maybe use "view" consistently instead of pure/impure/purity annotations, here, and in the rest of the proposal:

```suggestion
Any function that does not have a `view` annotation will be considered a non-view function. 

Function types can also have a `view` annotations, to be placed after the opening parenthesis but before the parameter list. So, for example, these are valid types:
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+    }
+    ```
+
+Any functions declared to be `view` will enforce that none of these operations occur in their bodies. If such an operation does occur, then the checker will report
+an error identifying the impure operation. Impure function declarations have no additional typechecking behavior; any operations are permitted.
+
+Note that the current design allows both emitting events and logging data from within a view function. These are traditionally considered side effects, but 
+it is perhaps worth some discussion on whether or not these should be allowed in `view` functions.  
+
+Purity also interacts covariantly with function subtyping: `view` functions are a subtype of impure functions. So, the following declarations would typecheck:
```
turbolent, line 139: As mentioned above, this also applies to interface and nested type requirement conformance, so maybe add this to the proposal and provide a similar example.

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+
+```cadence
+    let a: (view (): Void) = view fun() {}
+    let b: ((): Void) = view fun() {}
+    let c: ((): Void) = fun() {}
+    let d: (((view (): Void)): Void) = fun foo(x:((): Void)) {} // contravariance
```
turbolent, line 145: It looks like the extra parentheses for the first parameter of the function type are not necessary.
Maybe also make the comment more descriptive:

```suggestion
    let d: ((view (): Void): Void) = fun foo(f: ((): Void)) {} // function parameters are contravariant
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+while these would not: 
+
+
+```cadence
+    let x: (view (): Void) = fun() {}
+    let y: ((((): Void)): Void) = fun foo(f:(view (): Void)) {} // contravariance
```
turbolent, line 153:
```suggestion
    let y: (((): Void): Void) = fun foo(f: (view (): Void)) {} // function parameters are contravariant
```

flips/20220715-cadence-purity-analysis.md
```diff
@@ -0,0 +1,211 @@
...
+```
+
+Once purity is enforced in functions with `view` annotations, in order to require it in function conditions we can 
+simply treat pre-conditions and post-conditions as view contexts as if they had `view` annotations themselves. 
+
+### Drawbacks
+
+This adds additional complexity to the langauge; users will need to think about the purity of their functions and how they interact with each other in order 
+to use the new `view` annotations. However, these annotations are unlikely to be frequently required; they will only be necessary for functions that are 
+called from conditions (and functions called by those functions). Users will not need to interact with this system most of the time.
```
turbolent, line 163:
```suggestion
The proposed feature adds additional complexity to the language; 
users will need to think about the side-effects of their functions and how they interact with each other in order  to use the new `view` annotation.
However, this annotation is unlikely to be frequently required; it will only be necessary for functions that are 
called from conditions (and functions called by those functions). 
Users will not need to interact with this language feature most of the time. 
```

turbolent (approved):
This will be a great addition to the language 👍

flips/20220715-cadence-purity-analysis.md
dsainati1: Yes; the implementation already does this!

PR #1071, opened 2022-07-25
FLIP: Cadence borrowContract - Allow Dynamic Import

bluesign:
FLIP for new publicAccount method to Cadence to borrow contracts dynamically.  

______

For contributor use:

### DIFF
--- a/flips/20220725-cadence-borrowContract.md
+++ b/flips/20220725-cadence-borrowContract.md
@@ -1,0 +1,75 @@
+# borrowContract function for PublicAccount
+
+| Status        | Proposed    												  |
+:-------------- |:------------------------------------------------------------|
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN)  	 		  |
+| **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)  					  |
+|				| Austin Kline (austin@flowty.io)  							  | 
+| **Sponsor**   | 														      |
+| **Updated**   | 2022-07-25                                    			  |
+
+## Objective
+
+This proposed change will allow dynamic import of contracts from an address, by
+using a well known `borrow` dynamic.
+
+## Motivation
+
+There is a popular request, especially from people coming from EVM space, to
+have a way to import contracts from an address by name ( dynamic import
+functionality )
+
+Cadence lacks with dynamic imports, which leads to mass imports in some 
+contracts ( e.g. alchemy, find ) which also causes a lot of contracts to fail, 
+if one of the important contracts cease to exist. 
+
+## User Benefit
+
+- It makes it possible to use patterns that are common to Solidity developers 
+in Flow. 
+- It allows importing and using Contracts with name collision https://github.com
+/onflow/cadence/issues/1171
+- It will prevent single contract breaking the ecosystem
+
+## Design Proposal
+
+We suggest to add an new method to `PublicAccount` called `borrowContract`.
+
+```cadence 
+fun borrowContract<T: &Any>(name: String): T?
+```
+
+Exmaple usage will be:
+
+```cadence 
+if let contract = getAccount(0x1).borrowContract<&AnyResource{NonFungibleToken}>("ExampleNFT") {
+	let emptyCollection <- contract.createEmptyCollection()
+}
+```
+
+Another alternative is to add `borrow` method to `DeployedContract`:
+
+```cadence 
+fun borrow<T: &Any>(): T?
+```
+
+Exmaple usage will be:
+
+```cadence 
+if let contract = getAccount(0x1).contracts.get("ExampleNFT").borrow<&AnyResource{NonFungibleToken}>() {
+	let emptyCollection <- contract.createEmptyCollection()
+}
+```
+
+### Compatibility
+
+This should be backwards compatible. 
+
+### User Impact
+
+This will change the way users write their smart contracts. This can cause some issues with reading contract code due to the nature of dynamic imports. But eventually we don't see any security impact. 
+
+
+## Prior Art
+
+Solidity has a way to import and call contracts from address. 


flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+
+Exmaple usage will be:
+
+```cadence 
+if let contract = getAccount(0x1).borrowContract<&AnyResource{NonFungibleToken}>("ExampleNFT") {
+	let emptyCollection <- contract.createEmptyCollection()
+}
+```
+
+Another alternative is to add `borrow` method to `DeployedContract`:
```
dsainati1, line 50: The FLIP should propose one of these two and put the other in the "Alternatives" section. As it is now I am not sure which of the two specifically you are proposing.

dsainati1 (commented):
I am unsure about how I feel about this proposal. There are some significant advantages to being able to analyze which contracts are being imported statically; it allows us to create a contract dependency graph, for example, which we can use to optimize things like program caches to improve performance. Would it be reasonable to restrict the string argument to `borrowContract` to being a string literal so we can still do static analysis? Or would that defeat the point of the feature? If `borrowContract` could accept any string-typed expression I would need to see a very compelling case that the benefit here outweighs the cost.

dsainati1:
> But I think program cache argument is a good one. Do you have any idea how will it look when contract dependency graph is used?

Yeah. Currently the program cache is completely wiped on a contract update, because we don't know which downstream contracts might be affected by this update. If we have a dependency graph though, once update we are considering is to use this information to only invalidate cached contracts that depend (transitively) on the updated contract; which should dramatically decrease the amount of unnecessary cache invalidation that we do. 

A solution to https://github.com/onflow/cadence/issues/1684 would also likely require a dependency graph, as the way to avoid double counting interactions during imports would require us to know "where" interactions came from when they are imported. 

These are just the examples that I know of off the top of my head, but I believe @SupunS and @ramtinms have been thinking about this as well.

bluesign:
>  If we have a dependency graph though 

@dsainati1  I think you mean dependency graph of all contracts on chain, not the contract got updated right?

bluesign:
> We would only need to know which contracts are downstream of the one that was updated

So we only care about the descendants of the contract. ( contracts that are importing mine ) 

Let's call this example contract: `0xMAIN.Test`, if I write below code: 

```cadence 
import NonFungibleToken from 0xSOMEWHERE 
pub fun someFunction(){
  if let contract = getAccount(0x1).borrowContract<&AnyResource{NonFungibleToken}>("ExampleNFT") {
 	  let emptyCollection <- contract.createEmptyCollection()
   }
}
```

That means: 
- 0x1.ExampleNFT imports 0xSOMEWHERE.NonFungibleToken
- 0xMAIN.test imports 0xSOMEWHERE.NonFungibleToken

Let's say we have 3 cache slots: 

- 0x1.ExampleNFT
- 0xSOMEWHERE.NonFungibleToken
- 0xMAIN.test 

When a transaction execute: 

```cadence
import test  from 0xMAIN
transaction{
  execute(){
    test.someFunction()
  }
}
```

Now order of business here : 

- getProgram(0xMAIN.test) //cached. ( I think which also includes NonFungibleToken here ) 
- getProgram(0x1.ExampleNFT) //cached


this would not be no different than:

import NonFungibleToken from 0xSOMEWHERE 
pub fun someFunction(@NonFungibleToken.NFT){
  //where NonFungibleToken.NFT is type 0x1.ExampleNFT.NFT
}
```

@dsainati1  I don't know much about internals,  can you show me what I am missing here ?

**Comment by SupunS** [comment]
I'm yet to fully go through the FLIP. But as for the program cache problem, I think statically knowing the dependency graph is not always required. (Feels like going a bit off topic here, maybe we should move that to a separate discussion/thread). 

For example, a static dependency graph is only required if we are going to do an 'active' cache invalidation (i.e: invalidate cache at the time of deployment). But for a 'passive' invalidation, it is possible to invalidate (re-parse/re-check) a contract only when that particular contract is used by someone.

e.g: consider the scenario where `A` imports `B`. And `B` is updated.
- Active mode: 
  - Option I: Wipe the entire cache (This is how it happens at the moment)
  - Option II: Clear `A`'s cache (and only A) when `B` is updated. (This is the future improvement @dsainati1 is pointing out that would require knowing who is importing who. I guess the argument there is, if `A` had dynamically imported `B`, then this option-II is not feasible)
- Passive mode (concept only):
  - Every contract in the cache has a hash/uuid.
  - `A` keeps the hash/uuid of the imported `B`
  - When `B` is updated, then `B`'s hash/uuid changes. No changes done to `A` (or `A`'s cache)
  - When someone imports `A`, check `A`'s imports for any hash changes. If there are changes, then invalidate `A`.
  - This approach does not need to know the dependents.

I'm not really sure about the other problem (https://github.com/onflow/cadence/issues/1684) though, but it might be possible to do something similar there as well.

File: flips/20220725-cadence-borrowContract.md [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+```
+
+### Compatibility
+
+This should be backwards compatible. 
+
+### User Impact
+
+This will change the way users write their smart contracts. This can cause some issues with reading contract code due to the nature of dynamic imports. But eventually we don't see any security impact. 
+
```
SupunS, line 71: There was another approach suggested and some concerns were discussed in https://github.com/onflow/cadence/issues/550. Could those be applicable to this approach as well?

Also, can that be considered as an alternative approach?

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+We suggest to add an new method to `PublicAccount` called `borrowContract`.
+
+```cadence 
+fun borrowContract<T: &Any>(name: String): T?
+```
+
+Exmaple usage will be:
+
+```cadence 
+if let contract = getAccount(0x1).borrowContract<&AnyResource{NonFungibleToken}>("ExampleNFT") {
```
SupunS, line 45: One of my concerns with this approach is that, it allows having multiple variables pointing to the same contract. And also makes contracts behave like other cadence 'values'.

I think this is something that has been avoided by the design of cadence: https://docs.onflow.org/cadence/language/contracts/
> Contracts are types. They are similar to composite types, but are stored differently than structs or resources and cannot be used as values, copied, or moved like resources or structs

bluesign: Yeah actually, I choose borrow exactly for this reason. ( multiple variables pointing to the same contract ) 

By using reference to the contract, we avoid copy and move technically.

flips/20220725-cadence-borrowContract.md
bluesign: I think this is currently possible. ( package manager approach ) 

Main problem there is PackageManagerContract.
- It has to import a lot of other contracts and will hit interaction limit. 
- Even single one of the imports vanish, all contract will fail

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+have a way to import contracts from an address by name ( dynamic import
+functionality )
+
+Cadence lacks with dynamic imports, which leads to mass imports in some 
+contracts ( e.g. alchemy, find ) which also causes a lot of contracts to fail, 
+if one of the important contracts cease to exist. 
+
+## User Benefit
+
+- It makes it possible to use patterns that are common to Solidity developers
```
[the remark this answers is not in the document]
bluesign: I gave one as a code example ( creating collection example )  but I will try to add some more examples to the FLIP.

bluesign:
@SupunS Very good example with passive invalidation, thanks. 

Btw I still think programs should be cached separately and linked at runtime. And program cache should be something bounded like LRU cache. 

But what I want to point out here, in our proposal, it is nothing that prevents cache. And it is no different than getting an object from storage ( which you haven't important their contract ) and operating on the interface. ( think this transaction as a function of a contract for example: https://github.com/onflow/flow-ft/blob/master/transactions/generic_transfer.cdc )

Maybe I am missing something here.

SupunS:
@bluesign Yeah, I do agree, caching should be a separate problem. i.e: internal details like caching shouldn't prevent Cadence language features. @dsainati1 also has a point, that there may be implementation complexities. But again, I think this whole discussion should happen elsewhere, perhaps a separate Cadence git issue.

I would very much like to get the focus back on the core idea/proposal of the FLIP.

bluesign:
@SupunS To prevent this kind of contracts is sufficient benefit alone for me: 

https://flowscan.org/account/0xeb8cb4c3157d5dac

austinkline:
> @dsainati1 @SupunS To prevent this kind of contracts is sufficient benefit alone for me:
> 
> https://flowscan.org/account/0xeb8cb4c3157d5dac

This very contract is the reason I started asking @bluesign about this issue. Over time these kinds of setups get so complex or import so much that they cannot even function anymore. If instead you could borrow a contract with an interface, Alchemy could have had contracts which wish to be supported implement that interface and then it's just a standard contract call.

One big issue that has happened with this setup in the past, and which @bluesign has flagged, is that a bad update to one contract bricks the whole thing because of its imports. When the June 15th secure cadence spork came around, it took a while to get this these contracts operational again. 

But it isn't just about when cadence has breaking changes, theoretically this setup means any single contract in the import lists of these contracts can hold the others hostage by making their contract invalid somehow until maintainers can update the contract. Dynamic imports would not only make this kind of contract simpler, it would insulate them all from each other.

flips/20220725-cadence-borrowContract.md
turbolent: That's a good idea. Could you please add this reasoning to the proposal itself?

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+- It allows importing and using Contracts with name collision https://github.com
+/onflow/cadence/issues/1171
+- It will prevent single contract breaking the ecosystem
+
+## Design Proposal
+
+We suggest to add an new method to `PublicAccount` called `borrowContract`.
+
+```cadence 
+fun borrowContract<T: &Any>(name: String): T?
```
turbolent, line 39: This seems related to the contracts functionality of an account, which is already grouped into ``PublicAccount.Contracts`. Also, this function should also be available for `AuthAccount`s, i.e. be added to `AuthAccount.Contracts`. Maybe also rename the function to just `borrow`, as the other functions don't contain the noun either (e.g. `get`, `add`, etc.)

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+Another alternative is to add `borrow` method to `DeployedContract`:
+
+```cadence 
+fun borrow<T: &Any>(): T?
+```
+
+Exmaple usage will be:
+
+```cadence 
+if let contract = getAccount(0x1).contracts.get("ExampleNFT").borrow<&AnyResource{NonFungibleToken}>() {
```
turbolent, line 59: Downside of this is that the whole contract information is fetched (e.g. code), then thrown away (only the name is used)

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
+# borrowContract function for PublicAccount
+
+| Status        | Proposed    												  |
+:-------------- |:------------------------------------------------------------|
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN)  	 		  |
```
turbolent, line 5:
```suggestion
| **FLIP #**    | [1071](https://github.com/onflow/flow/pull/1071)  	 		  |
```

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+| **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)  					  |
+|				| Austin Kline (austin@flowty.io)  							  | 
+| **Sponsor**   | 														      |
+| **Updated**   | 2022-07-25                                    			  |
+
+## Objective
+
+This proposed change will allow dynamic import of contracts from an address, by
+using a well known `borrow` dynamic.
```
turbolent, line 14: The objective should focus on the goal, not the *how*. Move the second part to the "Design Proposal" section

```suggestion
This proposal will allow Cadence programs to import contracts dynamically from an address.
```

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,75 @@
...
+Cadence lacks with dynamic imports, which leads to mass imports in some 
+contracts ( e.g. alchemy, find ) which also causes a lot of contracts to fail, 
+if one of the important contracts cease to exist. 
+
+## User Benefit
+
+- It makes it possible to use patterns that are common to Solidity developers 
+in Flow. 
+- It allows importing and using Contracts with name collision https://github.com
+/onflow/cadence/issues/1171
```
turbolent, line 31:
```suggestion
- It allows importing and using contracts with name collision. See https://github.com/onflow/cadence/issues/1171
```

flips/20220725-cadence-borrowContract.md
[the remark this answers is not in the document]
turbolent: A good use-case could maybe be to create an empty FT vault?

turbolent (approved):
Great idea! 

@bluesign would you maybe be able to implement this in Cadence?

turbolent:
Just read through the comments. 

Cadence intentionally includes features that enable interoperability (e.g. dynamic dispatch, interfaces, etc.), which compromise the ability to statically analyze programs. The proposed feature follows that direction.

We dropped the early idea of requiring transactions to statically specify all accounts which will be read from/written to upfront, because it hinders interoperability in an organic ecosystem. The same point can be made about requiring programs to statically specify all contracts they will interact with during execution.

flips/20220725-cadence-borrowContract.md
bluesign: @turbolent how about putting `borrow<&T>(): &T` method to DeployedContract ? I think I can make T somehow inferred automatically right?

flips/20220725-cadence-borrowContract.md
bluesign: ah I just see this comment after replying to the other one. I will try to implement borrow(contractName: String)

bluesign:
I implemented this ( I hope ), should I make a PR for it and then go over FLIP ? or go over FLIP and then make the PR ?

turbolent:
@bluesign Awesome! 👏 🏅 

Please feel free to open a PR for the implementation in the Cadence repo and then link to it here. Having an implementation for it greatly helps with evaluating the proposal (e.g. it demonstrates that what is being proposed can be implemented, and allows contributors and developers to try the feature and provide feedback).

bluesign:
I am pretty noob in this area but tried to make something working. I may have some huge misunderstanding or mistakes there.  https://github.com/onflow/cadence/pull/1934

flips/20220725-cadence-borrowContract.md
turbolent: It seems like the two features, borrowing a contract as proposed in this proposal, and dynamically resolving an import (onflow/cadence#550), are two related features, but not alternatives. It is maybe worth mentioning the related feature

flips/20220725-cadence-borrowContract.md
turbolent: Maybe move this alternative `borrow` function into a "Alternatives Considered" section, and propose just one API, the `borrowContract` function.

bjartek (approved):
I really like this feature!

joshuahannan (approved):
I think this will be very useful!

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,77 @@
...
+Borrowing a Contract will return a reference to the contract, avoiding move/copy problems. 
+
+Implemented a proof of concept in https://github.com/onflow/cadence/pull/1934
+
+```cadence 
+fun borrow<T: &Any>(name: String): &T?
+```
+
+```cadence 
+if let contract = getAccount(0x1).contracts.borrow("ExampleFT") {
```
turbolent, line 41: The function call requires a type argument for the type parameter `T`:

```suggestion
if let contract = getAccount(0x1).contracts.borrow<&FungibleToken>("ExampleFT") {
```

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,77 @@
...
+
+This change should be backward compatible. 
+
+## User Impact
+
+The proposed change can affect how users write their smart contracts, which can cause some minor issues with reading contract code due to the nature of dynamic imports. But eventually, we don't see any negative security impact. 
+
+## Prior Art & Related Features 
+
+- Dynamically resolving import addresses [https://github.com/onflow/cadence/issues/550]
```
turbolent, line 74:
```suggestion
- Dynamically resolving import addresses https://github.com/onflow/cadence/issues/550
```

dete:
Over time, it's starting to seem to me that the best equivalent to a "contract" in Solidity isn't a contract in Flow, but instead it is "Type".

The specific motivating example used in this example talks about creating empty Vaults or empty Collections, but I think that an even bigger problem in Cadence is that you might want to reference Types from contracts you haven't imported, and I don't believe this proposal solves that problem. (Imagine a generic transaction template for buying an NFT. That template should have a mechanism for checking the type and amount of the purchase price, and the type and ID of the NFT. To do that strictly with native Type objects would require that the import statements are dynamic, or that you import all possible FT/NFT types!)

Personally, I would rather see us creating a mechanism to reference a Type without an explicit import, and then have empty Vault creation be a static method on the Type object, instead of a method on the Contract object.

Java has a robust notion of referencing types without imports, the following two snippets are equivalent and equally valid:

```Java
import java.lang.Math

int big = Math.max(5, 10)
```

```Java
int big = java.lang.Math.max(5, 10)
```

Perhaps in Cadence we should have something like this:

```
import FungibleToken from 0xf233dcee88fe0abe

let ftType: Type<&FungibleToken> = Type<&FlowToken.Vault from 0x1654653399040a61>()

let emptyV = ftType.createEmptyVault()
```

bluesign:
@dete I really liked this; much more useful. Also, it is extending the FLIP instead of limiting it. 

Pros:

- We can move createEmptyVault and createEmptyCollection to Vault and NFT ( as static methods ) with upcoming NFTv2 and FTv2 this can solve a lot of problems. 
- I think we can also allow: `let ftType: Type<&FungibleToken from 0xf233dcee88fe0abe>`

Cons: 
- We need a way to implement static methods/fields (the `static` keyword can solve those ) But it requires updates to contracts, to make it backwards compatible need to set the contract method/fields default to static probably. 
- Probably we need shortcut methods for `CompositeType` and` ReferenceType` combined, for dynamic Type runtime generation. 


Also do we need reference type here in the syntax?, maybe something like:

````
let ftType: Type<@FungibleToken.Vault> = Type<@FlowToken.Vault from 0x1654653399040a61>()
```

flips/20220725-cadence-borrowContract.md
```diff
@@ -0,0 +1,77 @@
+# borrowContract function for PublicAccount
+
+| Status        | Proposed    												  |
```
turbolent, line 3:
```suggestion
| Status        | Approved    												  |
```

turbolent (approved):
No votes against, approved 🎉

bluesign:
I am starting documentation asap.

PR #1092, opened 2022-08-15
docs for observer nodes

vishalchangrani:
Closes: [#6336](https://github.com/dapperlabs/flow-go/issues/6336)

## Description

______

For contributor use:

### DIFF
--- a/docs/content/node-operation/full-observer-node.mdx
+++ b/docs/content/node-operation/full-observer-node.mdx
@@ -4,11 +4,31 @@ sidebar_title: Observer Node
 description: Running an observer node locally
 ---
 
+An observer node is similar to an access node and provides a locally accessible, continuously updated and verified copy of the block data. It serves the Access gRPC API but, unlike an access node, an observer node does not need to be staked and **anyone** can run it without being added to the approved list of nodes.
 
-An observer node is similar to an access node and provides a locally accessible, continuously updated and verified copy of the block data.
-It can be easily run on any consumer-grade computer which has enough disk space.
+The observer node bootstraps by connecting to an access node and becoming part of the public network comprised of access nodes and other observer nodes. It then continuously receives blocks that are being added to the chain either directly from the access node or from other observer nodes that are part of the public network.
+However, it makes no trust assumption of the upstream access node or the observer node which is providing the block and locally verifies that the blocks that are received are the correct extension of the chain e.g After receiving valid blocks A, B and C when it receives block D, it verifies that block D is indeed signed by the consensus nodes and is a valid next block. The block data that is received is indexed and made available via the Access gRPC API. For Collection, Transactions and Account queries, it delegates those requests to the upstream access node. Similarly, transactions and scripts sent to an observer node are also forwarded to the upstream access node. Future versions of the observer node will be able to serve this data locally as well.
+
+Since the observer node is not staked, it does not produce or execute blocks but instead serves as an _unstaked access node_ that can be easily run on any consumer-grade computer which has enough disk space.
+
+## Who should run an observer Node?
+
+The observer node provides an alternative to running an access node. Hence it is ideal for Dapps that need access to latest block data available locally e.g a wallet application that needs to track the latest block ID and height. In addition to that, access node operators who want to scale their access nodes endpoints geographically can spin up geographically dispersed observer nodes which can talk to their staked access node and to each other.
 
 ## Running a node
+
+### Hardware
+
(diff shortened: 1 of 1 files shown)


docs/content/node-operation/full-observer-node.mdx
```diff
@@ -1,38 +1,221 @@
...
-node for local queries without the need for staking.
+The observer node bootstraps by connecting to an access node and becoming part of the public network comprised of access nodes and other observer nodes. It then continuously receives blocks that are being added to the chain either directly from the access node or from other observer nodes that are part of the public network.
+However, it makes no trust assumption of the upstream access node or the observer node which is providing the block and locally verifies that the blocks that are received are the correct extension of the chain e.g After receiving valid blocks A, B and C when it receives block D, it verifies that block D is indeed signed by the consensus nodes and is a valid next block. The block data that is received is indexed and made available via the Access gRPC API. For Collection, Transactions and Account queries, it delegates those requests to the upstream access node. Similarly, transactions and scripts sent to an observer node are also forwarded to the upstream access node. Future versions of the observer node will be able to serve this data locally as well.
```
peterargue, line 10: > makes no trust assumption of the upstream access node

Observers still must get the genesis block from a trusted source. This statement is true if the observer uses the `root-protocol-state-snapshot.json` file from the beginning of the spork. If it uses an AN to get its current snapshot, it must trust that AN.

docs/content/node-operation/full-observer-node.mdx
```diff
@@ -1,38 +1,221 @@
...
+
+#### Observer for Flow Testnet
+
+```
+docker run --rm \
+ -v /var/flow/bootstrap:/bootstrap:ro  \
+ -v /var/flow/data:/data:rw \
+ --name flow_observer \
+ --network host  \
+ --gcr.io/flow-container-registry/observer:v0.27.1 \
```
danielholmes839, line 158: `gcr.io/flow-container-registry/observer:v0.27.1` remove `--`

docs/content/node-operation/full-observer-node.mdx
```diff
@@ -1,38 +1,221 @@
...
+```
+
+#### Observer for Flow Testnet
+
+```
+docker run --rm \
+ -v /var/flow/bootstrap:/bootstrap:ro  \
+ -v /var/flow/data:/data:rw \
+ --name flow_observer \
+ --network host  \
```
danielholmes839, line 157: I think it might be better to recommend people run with `-p 9000:9000` instead since `--network host` doesn't work on mac. There's also a [`--publish-all`](https://docs.docker.com/network/host/) that might work.

docs/content/node-operation/full-observer-node.mdx
```diff
@@ -1,38 +1,221 @@
...
+gpg: Signature made Wed Sep 15 11:34:33 2021 PDT
+gpg:                using ECDSA key 40CD95717AC463E61EE3B285B718CA310EDB542F
+gpg: Good signature from "Flow Team (Flow Full Observer node snapshot verification master key) <flow-signer@onflow.org>" [unknown]
+gpg: WARNING: This key is not certified with a trusted signature!
+gpg:          There is no indication that the signature belongs to the owner.
+Primary key fingerprint: 7D23 8D1A E6D3 2A71 8ECD  8611 CB52 64F7 FD4C DD27
+     Subkey fingerprint: 40CD 9571 7AC4 63E6 1EE3  B285 B718 CA31 0EDB 542F
+```
+
+Alternately, if you don't care about the blocks before the current block, you can request the current root-snapshot file via the flow-cli.
```
peterargue, line 113: We should note here that the genesis block information MUST be from a trusted source as it's the foundation for all block verification. So if they choose this method, they must use an AN they trust.

vishalchangrani: The flow-cli actually supports using the secure connection. I switched the command to use that so that it is guaranteed that the snapshot file the user gets is the real one.

docs/content/node-operation/full-observer-node.mdx
vishalchangrani: removed the `--network host` and added `-p` for all ports.

docs/content/node-operation/full-observer-node.mdx
vishalchangrani: yup, removed. thanks for catching that.

peterargue (approved):

PR #1097, opened 2022-08-17
FLIP: Cadence - Enable new fields on existing resource and struct definitions

austinkline:
FLIP for enabling the addition of new fields in existing resource and struct definitions

For contributor use:

- [ ] Targeted PR against master branch
- [ ] Re-reviewed Files changed in the Github PR explorer

### DIFF
--- a/flips/20220817-add-fields-to-existing-definitions.md
+++ b/flips/20220817-add-fields-to-existing-definitions.md
@@ -1,12 +1,12 @@
 # Allow new fields in deployed Resources and Structs
 
-| Status        | Proposed                                                  |
-|---------------|:--------------------------------------------------------- |
-| **FLIP #**    | [1097](https://github.com/onflow/flow/pull/1097)            |
-| **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)                    |
-|               | Austin Kline (austin@flowty.io)                           | 
-| **Sponsor**   |                                                           |
-| **Updated**   | 2022-08-17                                                |
+---
+status: proposed 
+flip: TBD
+authors: Deniz Mert Edincik (deniz@edincik.com), Austin Kline (austin@flowty.io)  
+sponsor: Bastian Mueller (bastian.mueller@flowfoundation.org)
+updated: 2022-08-17
+---
 
 ## Objective
 This proposed change will allow existing structs and resources to add new fields to them by
--- a/flips/20220817-add-fields-to-existing-definitions.md
+++ b/flips/20220817-add-fields-to-existing-definitions.md
@@ -1,0 +1,80 @@
+# Allow new fields in deployed Resources and Structs
+
+| Status        | Proposed                                                  |
+|---------------|:--------------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN)            |
+| **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)                    |
+|               | Austin Kline (austin@flowty.io)                           | 
+| **Sponsor**   |                                                           |
+| **Updated**   | 2022-08-17                                                |
+
+## Objective
+This proposed change will allow existing structs and resources to add new fields to them by
+using optional and default assignment embedded in the object definition itself.
+
+## Motivation
+
+One major challenge in smart contract design in cadence is the inability to add new fields to already deployed 
+structs and resources. Not having the ability to do this means that contracts either get extended 
+with data that exists in another contract, entirely new contracts are made that are essentially reprints of the
+original, except with the newly added fields, or some other workaround such as storing fields as dictionaries
+and building them into at runtime.
+
+Each of these workarounds come with their downsides. 
+- New contracts lead to complex migrations for applications and sunsetting existing contracts. 
+- Hosting new data in another contract leads to harder-to-follow code and added complexity. 
+- Factory patterns increase compute since objects must be built at runtime and also lose the benefits of type-checking
+    since the underlying structure is not truly typed.
+
+## User Benefit
+
+- It will allow developers to add fields as they need them, making contract development more focused on the needs of the current version
+    as opposed to undue complexity to take future plans into account.
+- It will allow more maintainable contract code with less splintered logic as existing contracts update themselves
+
+## Design Proposal
+
+We propose to add the ability to define new optional fields, and the assignment of default values to fields in
+structs and resources. All fields added after the first deployment of a contract must be either optional or have
+a default parameter assigned so that initialization of existing instances of these objects can be properly
+initialized.
+
+```cadence
+pub struct Message {
+    pub let content: String
+}
+```
+
+We could now add a new field to this struct in a few ways
+
+```cadence
+pub struct Message {
+    pub let content: String
+    
+    // new fields
+    // an optional new field. Existing instances of Message will have timestamp set to nil when accessed
+    // unless they are set at a later date
+    pub let timestamp: UInt64?
+    // a default initialized field, Existing instances of Message will take the default value.
+    pub let received: Bool = false 
+}
+```
+
+### Limitations
+
+- This will not allow existing fields to be altered. That is, you cannot take a field and alter its type.
+- This will not allow new fields to be derived from existing ones.
+
+### Compatibility
+
+This should be backwards compatible
+
+### User Impact
+
+- Cadence developers will be able to modify their contract more to their needs instead of over-designing with
+the first launch of their contract(s). 
+- If new unforeseen features or fixes require new fields, those additions can be kept in the original contract instead of being silo'd off into their own contract.
+
+## Prior Art
+
+N/A
--- a/flips/20220817-add-fields-to-existing-definitions.md
+++ b/flips/20220817-add-fields-to-existing-definitions.md
@@ -2,7 +2,7 @@
 
 | Status        | Proposed                                                  |
 |---------------|:--------------------------------------------------------- |
-| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN)            |
+| **FLIP #**    | [1097](https://github.com/onflow/flow/pull/1097)            |
 | **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)                    |
 |               | Austin Kline (austin@flowty.io)                           | 
 | **Sponsor**   |                                                           |
@@ -36,7 +36,7 @@ Each of these workarounds come with their downsides.
 
 We propose to add the ability to define new optional fields, and the assignment of default values to fields in
 structs and resources. All fields added after the first deployment of a contract must be either optional or have
-a default parameter assigned so that initialization of existing instances of these objects can be properly
+a default value assigned so that initialization of existing instances of these objects can be properly
 initialized.
 
 ```cadence


dsainati1:
In the future, if you're submitting a Cadence FLIP, please add some people from the Cadence team (e.g. myself, @turbolent, @SupunS and @robert-e-davidson3) as reviewers; otherwise we don't get notifications about these. I wasn't aware this even existed until someone pointed out to me!

austinkline:
> In the future, if you're submitting a Cadence FLIP, please add some people from the Cadence team (e.g. myself, @turbolent, @SupunS and @robert-e-davidson3) as reviewers; otherwise we don't get notifications about these. I wasn't aware this even existed until someone pointed out to me!

ahhh good to know! Noted for the future :+1:

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+```cadence
+pub struct Message {
+    pub let content: String
+    
+    // new fields
+    // an optional new field. Existing instances of Message will have timestamp set to nil when accessed
+    // unless they are set at a later date
+    pub let timestamp: UInt64?
+    // a default initialized field, Existing instances of Message will take the default value.
+    pub let received: Bool = false
```
dsainati1, line 59: The design proposal needs more detail about how optional or default initialized fields would interact with the existing initialization requirements. Do these fields need to be handled in the `init` function, or is it possible to omit them and have these fields be `nil` or the default value implicitly on initialization?

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+```
+
+### Limitations
+
+- This will not allow existing fields to be altered. That is, you cannot take a field and alter its type.
+- This will not allow new fields to be derived from existing ones.
+
+### Compatibility
+
+This should be backwards compatible
```
dsainati1, line 70: cc @ramtinms if implemented this might have implications for potential work on composite type inlining

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+
+### User Impact
+
+- Cadence developers will be able to modify their contract more to their needs instead of over-designing with
+the first launch of their contract(s). 
+- If new unforeseen features or fixes require new fields, those additions can be kept in the original contract instead of being silo'd off into their own contract.
+
+## Prior Art
+
+N/A
```
dsainati1, line 80: We should add a Questions and Discussion section to talk about related topics to this FLIP. In particular, one that comes to mind is how this is going to interact with the potential addition of extensions to the language proposed here: https://github.com/onflow/flow/pull/1101. As that proposal stands now, we'd need some way of handling the case where a user updates their struct or resource with a new field or method that also exists in an extension.  It's also worth considering the extent to which these two proposals overlap in their use cases.

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
+# Allow new fields in deployed Resources and Structs
+
+| Status        | Proposed                                                  |
+|---------------|:--------------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN)            |
```
dsainati1, line 5:
```suggestion
| **FLIP #**    | [1097](https://github.com/onflow/flow/pull/1097)            |
```

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+| **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)                    |
+|               | Austin Kline (austin@flowty.io)                           | 
+| **Sponsor**   |                                                           |
+| **Updated**   | 2022-08-17                                                |
+
+## Objective
+This proposed change will allow existing structs and resources to add new fields to them by
```
dsainati1, line 12: Would it be beneficial to allow updating contracts with new methods as well? Or just fields?

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+    // unless they are set at a later date
+    pub let timestamp: UInt64?
+    // a default initialized field, Existing instances of Message will take the default value.
+    pub let received: Bool = false 
+}
+```
+
+### Limitations
+
+- This will not allow existing fields to be altered. That is, you cannot take a field and alter its type.
```
dsainati1, line 65: Users already have the ability to remove fields from existing composites, so giving them the ability to add new fields can implicitly allow changing a field's type if a user removes an existing field (that had type `T1`) in one update and then adds it back with a different type `T2` in another update. We'd need a specific solution for this case. One obvious answer is to simply fail a contract update if existing instances of a struct or resource have conflicting types with the field being added, but there are other potential solutions. Either way we should make it explicit in the proposal.

dsainati1 (commented):
Nice. I like the idea here; I agree this would make life easier for Cadence developers. The proposal could use a little more detail in some places though, specifically with regards to how the proposed changes would interact with existing behavior of the language.

flips/20220817-add-fields-to-existing-definitions.md
turbolent: The current design of only allowing field initialization in one place, the initializer, instead of the initializer or in the field declaration, was deliberate: There is only one place to look at when reading the code and it avoids many design questions related to expressions in field declarations: 
- When is the expression evaluated? Once, "statically" for all instances? For each instance, before the initializer?
- What is the evaluation order? Can a field refer to another?

Is the default value only used when migrating existing values, or is it also used for new values?

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+
+- It will allow developers to add fields as they need them, making contract development more focused on the needs of the current version
+    as opposed to undue complexity to take future plans into account.
+- It will allow more maintainable contract code with less splintered logic as existing contracts update themselves
+
+## Design Proposal
+
+We propose to add the ability to define new optional fields, and the assignment of default values to fields in
+structs and resources. All fields added after the first deployment of a contract must be either optional or have
+a default parameter assigned so that initialization of existing instances of these objects can be properly
```
turbolent, line 39:
```suggestion
a default value assigned so that initialization of existing instances of these objects can be properly
```

turbolent (commented):
This feature has been a goal ever since the early days of Cadence, thank you for opening a FLIP for it.

The main reason why this feature has not been implemented yet is that it is unclear *how* it should be implemented. Even though designs/proposals for language features usually do not discuss implementation details, it is an important factor when deciding if the proposal should be accepted. 

It would be great if the proposal could describe the behaviour of the necessary migration, in particular: When is the migration performed? Eagerly migrating all existing data at contract update-time is not tractable on-chain. For example, common relational database systems take multiple seconds, minutes, or even hours to migrate tables when adding new columns. The migration should probably occur lazily, when an existing value is accessed.

Another open question is how this feature interacts with the existing feature of removing fields. Currently, this is implemented by only updating the code and not actually removing the data (for the same reasons as mentioned above). Allowing addition would lead to effectively allowing field updates, i.e. changing the type of a field, by first removing the field, then adding it back with a different type.

austinkline:
> Nice. I like the idea here; I agree this would make life easier for Cadence developers. The proposal could use a little more detail in some places though, specifically with regards to how the proposed changes would interact with existing behavior of the language.

To make sure I can cover what is missing as explicitly as possible, it sounds like two main pieces:
1. How would default initialized values work? Sounds like this approach could go against some core design decisions in cadence so this might be better off getting scoped out.
2. How would cadence handle these new fields on existing instances of composite types?

austinkline:
@turbolent replying inline here but wanted to make sure your points are all addressed to be the best of my ability:

> The main reason why this feature has not been implemented yet is that it is unclear _how_ it should be implemented. Even though designs/proposals for language features usually do not discuss implementation details, it is an important factor when deciding if the proposal should be accepted.

Agreed and makes total sense, I will circle back to this FLIP and add more detail so we have more to go on in terms of discussion. 

> It would be great if the proposal could describe the behaviour of the necessary migration, in particular: When is the migration performed? Eagerly migrating all existing data at contract update-time is not tractable on-chain. For example, common relational database systems take multiple seconds, minutes, or even hours to migrate tables when adding new columns. The migration should probably occur lazily, when an existing value is accessed.

A few folks I've discussed this topic with have brought up migrations as well. It is why this FLIP doesn't allow reliance on other fields which would make evaluating them much more difficult when they don't "exist" yet under the hood. My take here is also to lazily evaluate as they are accessed for the first time. That could be when we access the entire object, or it could be when the new field itself is accessed for the first time

> Another open question is how this feature interacts with the existing feature of removing fields. Currently, this is implemented by only updating the code and not actually removing the data (for the same reasons as mentioned above). Allowing addition would lead to effectively allowing field updates, i.e. changing the type of a field, by first removing the field, then adding it back with a different type.

Yes, this was a good callout which I had neglected to consider but is certainly a core part of the problems in allowing updates like this. My current leaning is that since new fields would need to be nullable (or have a default), and conflicting type would need to be returned as null if it is found to be "incorrect" That is, if I have 

```cadence
struct Foo {
  amount: Int
}
```

and I then remove the field, and try to add a new one with the same name:

```cadence
struct Foo {
  amount: UInt64?
}
```

Then existing instances of Foo would return nil when the `amount` field is accessed. That value would still remain unless explicitly overridden in the scope of a transaction 

Alternatively, if we allowed default values, instead of `nil` we would assign the colliding field to that default value.

```cadence
struct Foo {
  amount: UInt64 = 100
}
```

The above would mean that all existing instances of `Foo` which have the old type for `amount` would read as `100`

austinkline (commented):
Thanks for the comments, I will take some time this weekend to elaborate on the FLIP per everyone's feedback and go from there

bluesign:
> In the future, if you're submitting a Cadence FLIP, please add some people from the Cadence team (e.g. myself, @turbolent, @SupunS and @robert-e-davidson3) as reviewers; otherwise we don't get notifications about these. I wasn't aware this even existed until someone pointed out to me!

Yeah this is big problem usually, though I think everyone from Dapper should follow @onflow/flow repository at least ( I am following almost all Flow & Dapper repositories, it is taking just 20 mins with my morning coffee, 80% of activity is being flow-go  )

bluesign:
> Allowing addition would lead to effectively allowing field updates, i.e. changing the type of a field, by first removing the field, then adding it back with a different type.

I think forbidding removal and allowing addition can be net improvement

austinkline:
> > Allowing addition would lead to effectively allowing field updates, i.e. changing the type of a field, by first removing the field, then adding it back with a different type.
> 
> I think forbidding removal and allowing addition can be net improvement
> 
> One problem I see from developer perspective, is adding fields as optional. I think it is a bit changing the meaning of Optional. Maybe we can allow adding synthetics fields ? They can lead to lazy migration. But also they have the problem of evaluation order.

Not sure I follow how this would change the meaning of optionals? Is it really any different to a cadence dev whether the field used to be nil or is nil, now? Or do you mean to flow itself and how it treats them?

Is synthetic right here, though? Once a new field is set it would be as real as anything else. Forcing them to be optional just helps ensure we don't need a migration strategy for existing data which I would say is a core requirement to this FLIP given

bluesign:
For me, also I think in the design of Cadence, optional variable means more than variable can be nil.  ( though I can be wrong, maybe @turbolent can clarify more ) 

But with my limited knowledge of optionals, actually they are property of the object also. ( some property of an object being optional, is also a property of the object ) 

so when we add a new property, it is being optional should depend on the object, not our technical limitations. Ofc we can decide this is a valid trade off, and add new fields as optionals. 

For me migration is not that scary to be honest, if we forbid removal.

robert-e-davidson3 (approved):
lgtm though if this cannot be resolved with the extensions FLIP, I prefer that FLIP to this one

flips/20220817-add-fields-to-existing-definitions.md
turbolent: Right, functions can already be added

flips/20220817-add-fields-to-existing-definitions.md
turbolent: Could we maybe simplify the proposal, by only keeping the first part of the proposal, adding new optional fields (which get initialized to nil), and removing adding non-optional fields with "default values"?

That way we could get at least some way of adding fields supported fairly quickly and easily.
Later we could propose adding non-optional fields in a follow-up proposal.

flips/20220817-add-fields-to-existing-definitions.md
turbolent: We could probably propose similar functionality as proposed for type removal in https://github.com/onflow/flips/pull/276: If we propose to allow adding new fields, the removal of fields should be required to "tombstone" a field, so it cannot be re-added. 

For example, this could be done through something like a `#removedField(fieldName)` pragma.

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
+# Allow new fields in deployed Resources and Structs
```
turbolent, line 1: We might also allow contracts?

flips/20220817-add-fields-to-existing-definitions.md
```diff
@@ -0,0 +1,80 @@
...
+
+| Status        | Proposed                                                  |
+|---------------|:--------------------------------------------------------- |
+| **FLIP #**    | [1097](https://github.com/onflow/flow/pull/1097)            |
+| **Author(s)** | Deniz Mert Edincik (deniz@edincik.com)                    |
+|               | Austin Kline (austin@flowty.io)                           | 
+| **Sponsor**   |                                                           |
+| **Updated**   | 2022-08-17                                                |
```
turbolent, line 9:
```suggestion
---
status: proposed 
flip: TBD
authors: Deniz Mert Edincik (deniz@edincik.com), Austin Kline (austin@flowty.io)  
sponsor: Bastian Mueller (bastian.mueller@flowfoundation.org)
updated: 2022-08-17
---
```

turbolent:
As we're finishing work on Cadence 1.0, we can finally come back to proposals like this. Thank you everyone for your patience!

I left some comments and suggestions, and am happy to both sponsor this proposal, as well as apply suggested edits and add remaining details. Does that sound good to you @bluesign  @austinkline?

Once we addressed the outstanding issues, we can schedule to vote on it in the next WG call

flips/20220817-add-fields-to-existing-definitions.md
SupunS: I like the idea of tomb-stoning. However there is a chance that people have already removed fields from composites without the pragma, and the data for those fields still exist. So adding a new field could still cause type-safety issues. Maybe we could cleanup the composite values as part of the migration, to remove fields that are not part of their corresponding type? (IDK if this is possible / how to do that, though)

turbolent:
Moved this to the FLIP repo: https://github.com/onflow/flips/pull/295

Let's continue the discussion there and hopefully we can decide on it in the upcoming WG call

PR #1122, opened 2022-09-12
FLIP for `publish` and `claim` for capabilities

dsainati1:
Adds a new proposal for a pair of API functions: `publish` and `claim` that are designed to address the capability bootstrapping use case detailed here https://github.com/onflow/cadence/issues/1951

______

For contributor use:

### DIFF
--- a/flips/20220908-publish-claim-capability.md
+++ b/flips/20220908-publish-claim-capability.md
@@ -113,13 +113,13 @@ the program will fail.
 This FLIP also adds two new events to Cadence:
 
 ```cadence
-event InboxValuePublished(provider: Address, name: String, type: Type) 
+event InboxValuePublished(provider: Address, recipient: Address, name: String, type: Type) 
 
 event InboxValueRemoved(provider: Address, remover: Address, name: String)
 ```
 
 These events are emitted whenever a value is added or removed from an inbox's publishing dictionary. When `publish` is called,
-`InboxValuePublished` is emitted containing the address of the `provider`, the `name` the value was published with, and the runtime
+`InboxValuePublished` is emitted containing the address of the `provider` and the `recipient`, the `name` the value was published with, and the runtime
 `type` it was published with. When a value is `claim`ed or `unpublish`ed, `InboxValueRemoved` is emitted, including the address the value was
 originally published from (`provider`), the address of the `remover` (in the case of `claim`, this is the recipient's address, in the case of 
 `unpublish` it is the same as the `provider`), and the `name` of the removed event. 
--- a/flips/20220908-publish-claim-capability.md
+++ b/flips/20220908-publish-claim-capability.md
@@ -5,7 +5,7 @@
 | **FLIP #**    | [1122](https://github.com/onflow/flow/pull/1122)     |
 | **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
 | **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
-| **Updated**   | 2022-09-14                                           |
+| **Updated**   | 2022-09-16                                           |
 | **Obsoletes** | https://github.com/onflow/flow/pull/945              |
 
 ## Objective
@@ -49,11 +49,11 @@ struct Inbox {
 
     fun unpermit(_ provider: Address)
 
-    fun publish(_ value: Any, name: String, recipient: Address): Bool
+    fun publish(_ value: Capability<&Any>, name: String, recipient: Address): Bool
 
-    fun unpublish<T : Any>(_ name: String): T?
+    fun unpublish<T : &Any>(_ name: String): Capability<T>?
 
-    fun claim<T: Any>(_ name: String, provider: Address): T?
+    fun claim<T: &Any>(_ name: String, provider: Address): Capability<T>?
 }
 ```
 
@@ -82,7 +82,7 @@ publishing dictionary for the `name` key. If the key does not exist on the map, 
 If the key does exist, then `claim` compares the stored `recipient` (from the original call to `publish`) 
 to see if it matches the address of the account calling `claim`. If it does not, then `claim` returns `nil`. 
 If it does match, then the runtime type of the published `value` compared against the type argument `T`. If 
-it does not match, then `claim` returns `nil`. If it does match, `value` is removed from the `provider`'s dictionary and 
+it does not match, then `claim` will produce an error. If it does match, `value` is removed from the `provider`'s dictionary and 
 returned to the `claim` calling account. In effect, this means that a `publish`ed value can only be `claim`ed once. 
 
 An example of how this might look, when bootstrapping a capability to a resource owned by 0x1:
@@ -123,15 +123,14 @@ stops taking up space if it goes un`claim`ed by its intended recipient. The call
 of the value) calls `unpublish` with the same `name` that the value was originally stored with. If a value with that 
 `name` is present in the account's publishing dictionary, and the provided type argument to `unpublish` is a supertype
 of that value's runtime type, then the function will return that value and remove it from the dictionary. Otherwise, 
-the function returns `nil`. 
+the function returns `nil`. The type comparison is a force-cast like `load` and `claim`, so if the types do not match
+the program will fail. 
 
 ### Alternatives Considered
 
-* The largest point of considering is whether or not `publish` and `claim` should be generalized
-to work with all values (as in the current proposal), or whether they should be limited to only 
-publishing/claiming capabilities. While the generalized versions are strictly more powerful, if we
-want to encourage users to use capabilities to objects rather than directly sending them back and forth, 
-we may not want to have such powerful alternatives available. 
+* The original proposal for this flip generalized `publish` and `claim` to work with resources, as well as Capabilities.
+While the generalized version was strictly more powerful, we felt that it was too likely to encourage users
+to write code in the `msg.sender` pattern instead of using it for the capability bootstrapping usecase.
 
 ### Best Practices
 
@@ -149,9 +148,3 @@ to add empty dictionaries to all accounts during the spork that adds support for
 ### User Impact
 
 * This should have no direct impact on users; it will not break any existing contracts. 
-## Questions and Discussion Topics
-
-* The API currently provides a way for users to `claim` values that have been published for them, 
-but once a value is `publish`ed, the only way to remove it from the account that published it is for
-the intended recipient to `claim` it. Is it worth having an `unpublish` function that will remove the
-entry for the value?
--- a/flips/20220908-publish-claim-capability.md
+++ b/flips/20220908-publish-claim-capability.md
@@ -2,17 +2,17 @@
 
 | Status        | (Proposed / Rejected / Accepted / Implemented)       |
 :-------------- |:---------------------------------------------------- |
-| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
+| **FLIP #**    | [1122](https://github.com/onflow/flow/pull/1122)     |
 | **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
 | **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
-| **Updated**   | 2022-09-13                                           |
+| **Updated**   | 2022-09-14                                           |
 | **Obsoletes** | https://github.com/onflow/flow/pull/945              |
 
 ## Objective
 
 This adds two new functions that allow users to publish values for specific other
 users to claim, while ensuring that other users cannot claim these values. In particular, 
-this dramatically simplifies the use case for bootstraping capabilities. 
+this dramatically simplifies the use case for bootstrapping capabilities. 
 
 ## Motivation
 
@@ -38,22 +38,41 @@ to share specific values directly with another account.
 
 ## Design Proposal
 
-This adds three new functions to `AuthAccount`s:
+This adds a new field `inbox` to `AuthAccount`, that has a new `Inbox` struct type
+defined as a nested composite on `AuthAccount`:
 
 ```cadence
-fun publish(_ value: Any, name: String, recipient: Address)
+struct Inbox {
+    let allowlist: [Address]
 
-fun unpublish<T : Any>(_ name: String): T?
+    fun permit(_ provider: Address)
 
-fun claim<T: Any>(_ name: String, provider: Address): T?
+    fun unpermit(_ provider: Address)
+
+    fun publish(_ value: Any, name: String, recipient: Address): Bool
+
+    fun unpublish<T : Any>(_ name: String): T?
+
+    fun claim<T: Any>(_ name: String, provider: Address): T?
+}
 ```
+
+The `inbox` tracks an `allowlist` that keeps track of all the addresses that are allowed to 
+`publish` values to it. An recipient account can `permit` a provider account to send values to its `inbox`
+by calling `permit`, and revoke that permission using `unpermit`. 
+
+On `PublicAccount`, the `Inbox` type only contains the `allowlist` so that a prospective `provider` that would like
+to `publish` a value to a `recipient` can inspect the `recipient` account's `allowlist` to check whether they are
+`permit`ted to do so by checking whether the `provider`'s address appears in the `recipient`'s `allowlist`.
 
 The `publish` function takes a `value` argument, a `name` string that identifies it, 
 and a `recipient` address that specifies which account should be allowed to `claim` the 
 published `value` later. When `publish` is called, `value` and its intended `recipient` are stored
 in a publishing dictionary on the calling account (not accessible to users), with the `name` as its key.
 Note that this means any values that have been `publish`ed but have yet to be `claim`ed will count towards
-the `publish`ing account's storage usage. 
+the `publish`ing account's storage usage. An account can only `publish` values to other accounts that have
+`permit`ted it to do so; if the calling account does not have permission to `publish` a value to the `recipient`, 
+`publish` will return `false`. Otherwise, it will return `true`. 
 
 The `claim` function takes the `name` of the value to be claimed and an `provider` address that
 specifies which account is providing the value, as well as a type argument `T` that specifies
@@ -79,7 +98,7 @@ transaction() {
     // not involve linking to a concrete path, but at the moment the only way to create
     // a capability is via `link` 
     let cap = acct.link<&MyIntf>(/private/myCapability, target: /storage/myResource)
-    acct.publish(cap, name: "yourCapability", recipient: 0x2)
+    acct.inbox.publish(cap, name: "yourCapability", recipient: 0x2)
   }
 }
 ```
@@ -90,7 +109,7 @@ import MyIntf from 0x1
 
 transaction() {
   prepare(acct: AuthAccount) {
-    let cap = acct.claim<Capability<MyIntf>>("yourCapability", provider: 0x1)
+    let cap = acct.inbox.claim<Capability<MyIntf>>("yourCapability", provider: 0x1)
     // if we successfully obtain the capability, store it on our account
     if cap != nil {
         acct.save(cap, to: /storage/myCapability)
--- a/flips/20220908-publish-claim-capability.md
+++ b/flips/20220908-publish-claim-capability.md
@@ -5,7 +5,7 @@
 | **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
 | **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
 | **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
-| **Updated**   | 2022-09-12                                           |
+| **Updated**   | 2022-09-13                                           |
 | **Obsoletes** | https://github.com/onflow/flow/pull/945              |
 
 ## Objective
@@ -38,10 +38,12 @@ to share specific values directly with another account.
 
 ## Design Proposal
 
-This adds two new functions to `AuthAccount`s:
+This adds three new functions to `AuthAccount`s:
 
 ```cadence
 fun publish(_ value: Any, name: String, recipient: Address)
+
+fun unpublish<T : Any>(_ name: String): T?
 
 fun claim<T: Any>(_ name: String, provider: Address): T?
 ```
@@ -96,6 +98,13 @@ transaction() {
   }
 }
 ```
+
+The `unpublish` function exists so that a provider can remove a published value from their storage, so that it
+stops taking up space if it goes un`claim`ed by its intended recipient. The calling account (the original provider
+of the value) calls `unpublish` with the same `name` that the value was originally stored with. If a value with that 
+`name` is present in the account's publishing dictionary, and the provided type argument to `unpublish` is a supertype
+of that value's runtime type, then the function will return that value and remove it from the dictionary. Otherwise, 
+the function returns `nil`. 
 
 ### Alternatives Considered
 
--- a/flips/20220908-publish-claim-capability.md
+++ b/flips/20220908-publish-claim-capability.md
@@ -1,0 +1,169 @@
+# Publish/Claim Capability
+
+| Status        | (Proposed / Rejected / Accepted / Implemented)       |
+:-------------- |:---------------------------------------------------- |
+| **FLIP #**    | [NNN](https://github.com/onflow/flow/pull/NNN) (update when you have PR #)|
+| **Author(s)** | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Sponsor**   | Daniel Sainati (daniel.sainati@dapperlabs.com)       |
+| **Updated**   | 2022-09-09                                           |
+| **Obsoletes** | https://github.com/onflow/flow/pull/945              |
+
+## Objective
+
+What are we doing and why? What problem will this solve? What are the goals and
+non-goals? This is your executive summary; keep it short, elaborate below.
+
+## Motivation
+
+Why is this a valuable problem to solve? What background information is needed
+to show how this design addresses the problem?
+
+Which users are affected by the problem? Why is it a problem? What data supports
+this? What related work exists?
+
+## User Benefit
+
+How will users (or other contributors) benefit from this work? What would be the
+headline in the release notes or blog post?
+
+## Design Proposal
+
+This adds two new functions to `AuthAccount`s:
+
+```cadence
+fun publish(_ value: Any, name: String, recipient: Address)
+
+fun claim<T: Any>(_ name: String, provider: Address): T?
+```
+
+The `publish` function takes a `value` argument, a `name` string that identifies it, 
+and a `recipient` address that specifies which account should be allowed to `claim` the 
+published `value` later. When `publish` is called, `value` and its intended `recipient` are stored
+in a publishing dictionary on the calling account (not accessible to users), with the `name` as its key.
+
+The `claim` function takes the `name` of the value to be claimed and an `provider` address that
+specifies which account is providing the value, as well as a type argument `T` that specifies
+the type with which the value should be claimed. When called, `claim` will search the `provider`s 
+publishing dictionary for the `name` key. If the key does not exist on the map, `claim` returns `nil`. 
+
+If the key does exist, then `claim` compares the stored `recipient` (from the original call to `publish`) 
+to see if it matches the address of the account calling `claim`. If it does not, then `claim` returns `nil`. 
+If it does match, then the runtime type of the published `value` compared against the type argument `T`. If 
+it does not match, then `claim` returns `nil`. If it does, `value` is removed from the `provider`'s dictionary and 
+returned to the `claim` calling account. In effect, this means that a `publish`ed value can only be `claim`ed once. 
+
+An example of how this might look, when bootstrapping a capability to a resource owned by 0x1:
+
+```cadence
+// transaction 1 (from the original resource owner 0x1)
+import MyIntf from 0x1
+
+transaction() {
+  prepare(acct: AuthAccount) {
+    // create a capability to a resource stored at /storage/myResource.
+    // With the changes proposed in https://github.com/onflow/flow/pull/798, this may
+    // not involve linking to a concrete path, but at the moment the only way to create
+    // a capability is via `link` 
+    let cap = acct.link<&MyIntf>(/private/myCapability, target: /storage/myResource)
+    acct.publish(cap, name: "yourCapability", recipient: 0x2)
+  }
+}
+```
+
+```cadence
+// transaction 2 (from the reciever of the capability 0x2)
+import MyIntf from 0x1
+
+transaction() {
+  prepare(acct: AuthAccount) {
+    let cap = acct.claim<Capability<MyIntf>>("yourCapability", provider: 0x1)
+    // if we successfully obtain the capability, store it on our account
+    if cap != nil {
+        acct.save(cap, to: /storage/myCapability)
+    }
+  }
+}
+```
+
+### Drawbacks
+
+Why should this *not* be done? What negative impact does it have? 
+
+### Alternatives Considered
+
+* Make sure to discuss the relative merits of alternatives to your proposal.
+
+### Performance Implications
+
+* Do you expect any (speed / memory)? How will you confirm?
+* There should be microbenchmarks. Are there?
+* There should be end-to-end tests and benchmarks. If there are not 
+(since this is still a design), how will you track that these will be created?
+
+### Dependencies
+
+* Dependencies: does this proposal add any new dependencies to Flow?
+* Dependent projects: are there other areas of Flow or things that use Flow 
+(Access API, Wallets, SDKs, etc.) that this affects? 
+How have you identified these dependencies and are you sure they are complete? 
+If there are dependencies, how are you managing those changes?
+
+### Engineering Impact
+
+* Do you expect changes to binary size / build time / test times?
+* Who will maintain this code? Is this code in its own buildable unit? 
+Can this code be tested in its own? 
+Is visibility suitably restricted to only a small API surface for others to use?
+
+### Best Practices
+
+* Does this proposal change best practices for some aspect of using/developing Flow? 
+How will these changes be communicated/enforced?
+
+### Tutorials and Examples
+
+* If design changes existing API or creates new ones, the design owner should create 
+end-to-end examples (ideally, a tutorial) which reflects how new feature will be used. 
+Some things to consider related to the tutorial:
+    - It should show the usage of the new feature in an end to end example 
+    (i.e. from the browser to the execution node). 
+    Many new features have unexpected effects in parts far away from the place of 
+    change that can be found by running through an end-to-end example.
+    - This should be written as if it is documentation of the new feature, 
+    i.e., consumable by a user, not a Flow contributor. 
+    - The code does not need to work (since the feature is not implemented yet) 
+    but the expectation is that the code does work before the feature can be merged. 
+
+### Compatibility
+
+* Does the design conform to the backwards & forwards compatibility [requirements](../docs/compatibility.md)?
+* How will this proposal interact with other parts of the Flow Ecosystem?
+    - How will it work with FCL?
+    - How will it work with the Emulator?
+    - How will it work with existing Flow SDKs?
+
+### User Impact
+
+* What are the user-facing changes? How will this feature be rolled out?
+
+## Related Issues
+
+What related issues do you consider out of scope for this proposal, 
+but could be addressed independently in the future?
+
+## Prior Art
+
+Does the proposed idea/feature exist in other systems and 
+what experience has their community had?
+
+This section is intended to encourage you as an author to think about the 
+lessons learned from other projects and provide readers of the proposal 
+with a fuller picture.
+
+It's fine if there is no prior art; your ideas are interesting regardless of 
+whether or not they are based on existing work.
+
+## Questions and Discussion Topics
+
+Seed this with open questions you require feedback on from the FLIP process. 
+What parts of the design still need to be defined?
--- a/flips/20220908-publish-claim-capability.md
+++ b/flips/20220908-publish-claim-capability.md
@@ -43,7 +43,7 @@ defined as a nested composite on `AuthAccount`:
 
 ```cadence
 struct Inbox {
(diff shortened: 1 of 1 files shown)


dsainati1:
cc @bjartek @bluesign as you were involved in the discussion on the original FLIP

bluesign:
Maybe allowing claim to both accounts can be good. ( if I publish to wrong address etc ) 

Also will this dictionary be available to user ? Some kind of seeing what I published, when I published can be good. If I published you something 1 month ago, if you didn't claim I want to be able to clean probably.

bluesign:
Also `fun claim<T: Any>(_ name: String, provider: Address): T?` is there a reason this is not something like: `getAccount(provider).claim<T:Any>(_ name: String)` 

I think it will be more familiar if we somehow use `getAccount`.  

PS: unless we will have some kind of listing for potential things I can claim. But then we are becoming full inbox.

flips/20220908-publish-claim-capability.md
```diff
@@ -0,0 +1,129 @@
...
+}
+```
+
+```cadence
+// transaction 2 (from the reciever of the capability 0x2)
+import MyIntf from 0x1
+
+transaction() {
+  prepare(acct: AuthAccount) {
+    let cap = acct.claim<Capability<MyIntf>>("yourCapability", provider: 0x1)
```
bjartek, line 91: Have you considered a conveniene method to save the claim to a path directly? As that is what most people would want to do?

dsainati1:
> Also `fun claim<T: Any>(_ name: String, provider: Address): T?` is there a reason this is not something like: `getAccount(provider).claim<T:Any>(_ name: String)`
> 
> I think it will be more familiar if we somehow use `getAccount`.
> 
> PS: unless we will have some kind of listing for potential things I can claim. But then we are becoming full inbox.

At the moment the account calling the `claim` function has to be the recipient account: `recipient.claim<T>(name, provider)`. If the provider is the account on which the `claim` function is called, then we have no argument that tells us which account is actually claiming the object.

bluesign:
> At the moment the account calling the claim function has to be the recipient account: recipient.claim<T>(name, provider). If the provider is the account on which the claim function is called, then we have no argument that tells us which account is actually claiming the object.

Thanks, makes sense

flips/20220908-publish-claim-capability.md
```diff
@@ -0,0 +1,129 @@
...
+```cadence
+fun publish(_ value: Any, name: String, recipient: Address)
+
+fun claim<T: Any>(_ name: String, provider: Address): T?
+```
+
+The `publish` function takes a `value` argument, a `name` string that identifies it, 
+and a `recipient` address that specifies which account should be allowed to `claim` the 
+published `value` later. When `publish` is called, `value` and its intended `recipient` are stored
+in a publishing dictionary on the calling account (not accessible to users), with the `name` as its key.
```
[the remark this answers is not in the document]
austinkline: It might also make sense to emit an event when something is initially added to be claimed as well? Or is that considered out of scope? Seems like something folks would want to track:
- When a new capability or resource is claimable
- When it is claimed
- If the claimable was cancelled

austinkline:
Would it make sense to have a third function to revoke/cancel a claimable item as well? Or would the assumption be that you would instead just destroy the underlying capability/resource being claimed which would mean it can't happen anyway

dsainati1:
Thanks for the quick comments on this! Based on the discussion so far, it seems like we need two more additional pieces of functionality here:

1) The ability to revoke a published object. This would only be possible if the calling account was the one who originally published it (obviously) and would return the resource/capability/value in question, while removing it from the publishing dictionary. 

2) More information about what is available to be claimed and by whom. In the interest of not becoming a full inbox, as @bluesign said, I am leaning towards having this information be mostly off-chain, perhaps using events like @austinkline suggested. We could emit an event when something is published containing its name and who it is for, an event when it is claimed, and an event if it is unpublished. This way users who can choose to submit transactions to claim objects based on these events, but the publishing dictionary is not inspectable on-chain at will, as this would break the abstraction that the `publish`/`claim` API is intended to provide.

bluesign:
In the events discussion, problem here is: this is privileged function ( claim ), so we cannot use it as a building block, enrich with contracts, etc. So events are for sure important, but also I think some kind of on-chain query would be also nice. At least some list of addresses, which I can claim something ( like a notification )

dsainati1:
I added an `unpublish` function that takes the name of the published value and removes it and returns it.

flips/20220908-publish-claim-capability.md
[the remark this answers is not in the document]
janezpodhostnik: > How do I know what I can claim onChain if I cannot access it?
I'm not sure how valid this question is in the context of bootstrapping. When **A** is trying to give a capability specifically to **B**, does **B** really need on-chain information to help **B** claim the capability? 

**A** probably doesn't need to package something for a random account that it doesn't have an off-chain connection to, or is there a use case I missing?

flips/20220908-publish-claim-capability.md
```diff
@@ -0,0 +1,138 @@
...
+The `claim` function takes the `name` of the value to be claimed and an `provider` address that
+specifies which account is providing the value, as well as a type argument `T` that specifies
+the type with which the value should be claimed. When called, `claim` will search the `provider`s 
+publishing dictionary for the `name` key. If the key does not exist on the map, `claim` returns `nil`. 
+
+If the key does exist, then `claim` compares the stored `recipient` (from the original call to `publish`) 
+to see if it matches the address of the account calling `claim`. If it does not, then `claim` returns `nil`. 
+If it does match, then the runtime type of the published `value` compared against the type argument `T`. If 
+it does not match, then `claim` returns `nil`. If it does match, `value` is removed from the `provider`'s dictionary and 
+returned to the `claim` calling account. In effect, this means that a `publish`ed value can only be `claim`ed once.
```
janezpodhostnik, line 67: > value can only be `claim`ed once

Whats the reason for this? Is there any issue with claiming the same capability multiple times. why not just leave it.

dsainati1: This would allow the API to support publishing/claiming resources easily. If we decide that we only want this to work for capabilities we could discuss removing this limitation.

flips/20220908-publish-claim-capability.md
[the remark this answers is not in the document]
janezpodhostnik: Another thing to add here is that if events need to be emitted, the publishing party can always write a wrapper function in a smart contract, that publishes and emits an event.

On the other hand, there is no way for the publisher to react when the value has been claimed. Maybe when publishing you could specify what happens when claiming? Some sort of callback. (that callback could emit an event or remove the published value)

katelynsills:
I'm very supportive of this FLIP! It's a major improvement over the identity object proposal, and it doesn't encourage the msg.sender pattern. However, I think it introduces a major issue with spam and other abuse.

### Spam and Abuse

In capability-based security, a lot of the security comes from the fact that everyone is disconnected from each other until they explicitly consent to connect (and that connection can be revoked.) This FLIP proposes to change that. Now everyone has a direct, irrevocable connection to everyone else. Granted, there are some protections: the Provider is paying all the storage costs for what they are sharing.

However, relying on storage costs alone to prevent spam and abuse will break as soon as the incentive for abuse > storage costs. For instance, imagine a motivated and coordinated harassment attack, where the Recipient keeps getting notifications where the string `name` is used to send threatening and abusive messages. Thus, there needs to be an easy way to revoke this connection between accounts.

### Solutions

An easy solution would be to add a `block` method to the AuthAccount, which takes a sender address. However, that's still putting the burden on the user to actively block, rather than allowing them to affirmatively consent to interaction with specific other accounts.

I wonder, since this FLIP is supposed to only be about solving the bootstrap problem, if the publish/claim API should be much more limited. Specifically, the Recipient could publish only one capability per Provider. And that capability would be of a particular type: a private inbox. In this case, there's no need for a name that can be used to transmit abusive messages from the general public.

Here's how I imagine [Dete's example use case](https://github.com/onflow/flow/pull/945#discussion_r954405266) would work:

> Dete wants to share the pause capability with Bjarte. Bjarte, knowing that Dete is unlikely to spam or abuse him, creates an inbox capability specifically for Dete that allows Dete to send Bjarte anything with no rate limits. Bjarte publishes this for Dete, and Dete claims it. Now Bjarte can receive capabilities and resources from Dete without having to actively claim. We could imagine in the future that Bjarte's inbox actually decides how the received things are named and stored in Bjarte's account, thus giving Dete an attenuated authority to Bjarte's storage, restricted to the things Dete has sent as well as other rules (like subfolders or resources within collections).

As another use case, let's imagine that rather than Dete, the sender is someone that Bjarte doesn't trust much:

> Bjarte would want to only allow the minimum access that is needed. He wants the ability to revoke, of course, but also perhaps it's a one time inbox so access is automatically cut off without having to send another transaction to revoke. 

### Complications and other thoughts

* Since the boundary between a Provider and a Recipient is not asynchronous, if a Provider creates a transaction that starts executing the Recipient's inbox rules, there's the possibility for reentrancy and other malicious behavior by the Recipient. Requiring a `claim` in a later transaction is actually nicely asynchronous.
* As @janezpodhostnik has pointed out, the inbox model, which is often used in the capability world, doesn't work very well if the recipient pays for storage, due to the ability to spam. Possible solution: the recipient needs to be able to specify rules that run prior to storing anything on their side and prior to any notifications.

At this point, I think we definitely need to solve the spam/abuse problem, and I would love to find a way to do that using the private inbox pattern, since that is the traditional capability pattern for this. However, the above complications might make that infeasible. Still, any step towards a private inbox model, especially allowing for attenuation (rules about who can send what when), might be a step in the right direction.

dsainati1:
@katelynsills thank you for the detailed comment! I hadn't considered the potential for spam and abuse when designing the API, so I appreciate that you pointed it out here. Based on my understanding of your analysis (and please correct me if I am wrong here!), the primary issue with the API currently is that it allows a Provider to send something to any Recipient, without the Recipient's explicit consent. 

One potential option would be to reverse the responsibility relationship for who initiates the connection: instead of `publish` and `claim`, where the Provider initiates by publishing a value and the Recipient chooses to claim it later, we could perhaps have something like `request` and `provide`/`fulfill` where a Recipient requests a value of a specific type from a Provider, and the Provider later fulfills that request by sending the value to the Recipient. However this may still have the potential for spam, just in the other direction, with a Recipient spamming a provider with spurious requests. 

Another solution to this would be the inverse of the `block` method you described; a `permit` or `allow` method that a potential Recipient can call to add a potential Provider to their allowlist. This way a Provider can only `publish` a value to a Recipient that has explicitly opted to allow them to do so, and the Recipient can revoke this permission at any time via a method that would remove the Provider from this allowlist. We could extend this further by allowing the Recipient to specify a type argument that specifies exactly what types of values they wish to receive from the Provider they are allowlisting, and potentially even an argument specifying how many values they wish to receive before automatically closing the connection. 

This seems like a limited version of the private inbox paradigm you are suggesting, that also addresses the additional complications you mentioned: because the Recipient must still explicitly `claim` things that the Provider `publish`es to them, it maintains the asynchronous nature of the connection and also requires the Provider to pay for storage until such time as the Recipient consents to receive the value they are being sent. 

Let me know your thoughts on this! It's very possible I have missed some more cases here.

turbolent:
@bluesign 

> Maybe allowing claim to both accounts (sender and receiver) can be good. 

Do you mean that the sender is able to claim on behalf of the receiver, so they could unpublish the capability, e.g. to clean up?

> Some kind of seeing what I published, when I published can be good. 

Agreed! Some form of on-chain introspection would be great. We already have it for some on-chain data (e.g. storage iteration) and are currently adding it for other data (e.g. account keys) where it is missing. 

> Also `fun claim<T: Any>(_ name: String, provider: Address): T?` is there a reason this is not something like: `getAccount(provider).claim<T:Any>(_ name: String)`
> 
> I think it will be more familiar if we somehow use `getAccount`.

The `claim` function needs to be defined on `AuthAccount`, not `PublicAccount`, so that the claim function can verify the identity. If the `claim` function would be defined on `PublicAccount`, anyone could claim the capability, as anyone can call `getAccount` to get the `PublicAccount`.

bluesign:
> Do you mean that the sender is able to claim on behalf of the receiver, so they could unpublish the capability, e.g. to clean up?

yeah but @dsainati1 already added `unpublish`, though I was just gonna comment on that I think `recipient` should be in that too. ( if I will send 100 people FLOAT for example, I cannot name them all unique probably )

turbolent:
@katelynsills 

Good point! At first I assumed that such a system, given that it is a mailbox, would be similar to e.g. snail and electronic mail, which are opt-in by default. However, given that this FLIP proposes a system for a very specific and limited purpose, bootstrapping capabilities, requiring explicit permission from the receiver for the sender to publish something seems reasonable 👍

turbolent:
@bluesign 

> > Do you mean that the sender is able to claim on behalf of the receiver, so they could unpublish the capability, e.g. to clean up?
> 
> yeah but @dsainati1 already added `unpublish`, though I was just gonna comment on that I think `recipient` should be in that too. ( if I will send 100 people FLOAT for example, I cannot name them all unique probably )

Ah, I didn't saw that yet. An `unpublish` function seems functionally equivalent to allowing claiming by the sender, but it seems slightly more explicit.

dsainati1:
> if I will send 100 people FLOAT for example, I cannot name them all unique probably

I don't think this is the intended use case for this feature; if you want to send people tokens, for example, we'd much rather encourage you to use the existing `Vault` capabilities. By requiring people to give unique names to values they send through this API we can encourage it to be limited to distinct, personal exchanges rather than being a venue for broadcasting many identical values to many people.

katelynsills:
@dsainati1, I really like the `permit` idea. In a way, it's effectively solving the time-window problem that co-signed transactions had while keeping the same security. With `permit`, both accounts still consent to the connection (the Recipient through `permit`, the Provider through `publish`) but can now do so asynchronously.

Regarding attenuation, you're right that we could add opt-in restrictions as part of the API (specify the type, number of messages etc.). But, I think we don't want to hard-code it as part of the API because that's not encouraging people to create their own capabilities with their own rules. Ideally, if I want a new kind of rule, I shouldn't need to request a change to the API that everyone uses; I can just write the code that controls my inbox.

Given that the spam issue is solved with `permit`, I think it's fine to move forward without tackling attenuation, but I think capability-based attenuation might be a good future goal.

bluesign:
>  However, given that this FLIP proposes a system for a very specific and limited purpose, bootstrapping capabilities...

Are we planning to limit this to capabilities ? If so maybe we can merge this with Capabilities FLIP.

dsainati1:
> Are we planning to limit this to capabilities ? If so maybe we can merge this with Capabilities FLIP.

I still think this is an open question; the API is perfectly functional for values of any type, so if we choose to restrict it it would be for the express purpose of limiting the use cases, not because it would not work for sending non-capability values. 

Either way though, I don't think it makes sense to merge the FLIPs. This API is related-to but separate-from the Capability Controllers FLIP because it's designed to work independently of whatever changes are decided upon there; regardless of how Capabilities may change as a result of that, this should still be able to send them between accounts with this API. Also the Capability Controllers would be a breaking change (and thus need to wait for Stable Cadence), while this would not, so keeping these separate would allow us to get this feature to users faster.

bluesign:
Sure but if we go `permit` path. We are just adding something which we can achieve with a contract already. 

Considering below flow: 
0x2:  permit `0x1` to `publish` me the `resource`
0x1: `publish` the resource to `0x2`
0x2: `claim` from `0x1` the `resource`

this is already possible with: 
0x2: set up account, put some resourceReceiver resource with function `saveCapability(Capability)`
0x2: set whitelist address: [0x1]
0x1: call `saveCapability(Capability)` with your capability (cap)  ( 0x2: check if cap.address is in whitelist  ) 

All power of this API is removing need for setup.

dsainati1:
I have made a couple changes to the proposal:

1) These functions are now grouped into an `Inbox` nested struct on the `Account` type to keep them organized 

2) I've added an `allowlist` field to `inbox` that tracks the list of Addresses that are permitted to `publish` to that account. An account can add to it using `permit` and remove from it using `unpermit`. A Provider account can only `publish` to a Recipient account if the Recipient's `allowlist` contains the Provider's Address. In order to allow users to check whether they are able to send values to a Recipient, the `allowlist` (and only the `allowlist`) is readable on `PublicAccount` objects.

bluesign:
Few last thoughts: 

** Capability or Any **

I think this system can solve more than "capability bootstrap".  As long as we leave signature as `Any` it will be used as such creative ways. So decision to limit capabilities or not, changes all dynamics here. 

** Practical Usage **

Let's confess, every transaction in any dapp now will have one more prepend line ( like initialization of collection etc ), which will be something like: `acct.Inbox.permit(Address(DAPP_ADDRES))` running. So there is no single chance you can disallow any dapp you are using. 

**Spam/Abuse case**

I think permit here is overprotective. Maybe I am too old fashioned, but I think we have 2 pieces at work here: `Publishing` the capability ( or resource ) and `Notifying` the user that there is something waiting for them. 

As long as I don't get any notification, someone sharing me something somewhere is no base for Spam/Abuse. Technically in our system what we call `inbox` is the notification part. So maybe like every app on my phone ( allowing notifications on messages from app ) or my DHL package ( alert on shipment notifications which I subscribe ) or my bank transfers ( alert me on transfer from X ) , better option is to able to permit notifications instead of permitting publishing. 

Also if we have some opt-in notification system, it can also be valuable to dapps in other ways.

katelynsills:
@bluesign, good point that one way to reduce abuse is to ensure that `publish` doesn't notify the Recipient by default. But, doesn't that require that `publish` should not emit an event by default, unless things that consume event logs (wallets, block explorers?) are tailored to the account's individual tastes of who they want to hear from and when? That seems tricky to pull off in a way that's easier than `permit` - either way, the account has to have an allowlist at some level, so why not at this level? Seems like you would want to nip spam/abuse in the bud as early as possible rather than filter at a later stage.

bluesign:
@katelynsills It is a bit trade-off, but in practical sense, nothing will change. 

If you disable notifications, you will not know there is something for you at my account, which is practically same as I am not sharing to you. 

If you allow my notifications, your wallet etc can show you I shared with you something. 

If you didn't allow my notifications, but somehow from some side-channel I reached you: there is no need for 3rd party to know I published you something anyway. Also this allows a semi-privacy mode. ( even though nothing private on blockchain )

katelynsills:
@dsainati1, re: whether we should allow Resources to  be `published`. I think we shouldn’t, and I only just realized why. 

As we know, `msg.sender` is bad because it makes safe, permissionless delegation impossible. There’s a list of accounts that can do X, and if I want to allow someone else to act on my behalf, they either can’t do X (not on the list), or I have to add them to the list (but maybe I don’t have admin authority over the list, and also the person I give access to can’t further delegate without themselves having admin authority to the list, and so on), or I have to give them access to my account as me. (Sidenote: maybe we should add this explanation to the documentation?)

Compare this to capability-based security: I can delegate easily by creating a capability restricted only to X and pass it to someone else.  And that other person can delegate even further in the same way.[1]

`publish/claim` is very much an ACL/msg.sender pattern: only a particular account can `claim` and delegation is impossible. For bootstrapping, this is a necessary evil - we have to start somewhere. 

But if this becomes the standard way of sending things between accounts, I think this is a problem because it teaches ACLs and not capabilities. So how can we guide users such that `publish/claim` is only used for the initial contact between two accounts, after which they should communicate by sending things to each other’s capabilities? Not allowing Resources seems essential for that. 

-----------
[1] For anyone who might be skeptical of importance of safe, permissionless delegation, imagine grocery delivery as an analogy to users receiving things from dapps. In real life, either we go to the grocery store and pick things up ourselves, or we get groceries delivered (delegation). But, unfortunately we either have to give the delivery person [a key](https://www.amazon.com/Amazon-Key-In-Garage-Delivery/b?ie=UTF8&node=21222091011) to our house (creepy! And dangerous!) or the delivery sits on our front porch waiting for us to come home. Wouldn’t it be nice if the food we ordered appeared in exactly the right spot in our pantry and fridge? With attenuated delegation, we can do that! We can give someone access that only allows them to place items in specific spots in our pantry/fridge, and doesn’t give any access to anything else in our house. And if the grocery delivery person calls out sick, the company can get another delivery person easily, without having to add the new person to any allowlist or give the company control over who gets in my house. This enables cooperation where it was impossible before - I might trust a random person to deliver my groceries even if I would not trust them in my house. 

You can imagine this same pattern of delegation expanding beyond delivery and across domains: delegating to a meal planner who selects the grocer and does the purchasing according to diet restrictions and a budget and gives the pantry access capability to the delivery company without you having to. Because there's no central authority deciding the permissions, the potential for innovation is endless.

Long story short - safe, permissionless delegation is very valuable, especially for a platform like Flow where you want to see a ton of economic activity unleashed.

bluesign:
> publish/claim is very much an ACL/msg.sender pattern: only a particular account can claim and delegation is impossible. 

But this is the whole point of publish/claim, so far: "one-on-one easy transfer with verified identity" 

> Wouldn’t it be nice if the food we ordered appeared in exactly the right spot in our pantry and fridge? 

It is 'trade offs' all the way down, I mean we can extend on that: Wouldn't be nice if delivery person didn't know what they are delivering, or where they are delivering, or when they are delivering ?

bjartek:
To me this is turning out to be just a little bit more convenience over a well established pattern of capability receivers. If I have to permit things first I could as well set up a CapabilityReceiver. So then I am not sure it is worth the effort personally.

dsainati1:
> re: whether we should allow Resources to be published. I think we shouldn’t, and I only just realized why.

@katelynsills this is a pretty compelling explanation; I'm convinced it makes sense to limit this API to only sending and receiving capabilities. 

> To me this is turning out to be just a little bit more convenience over a well established pattern of capability receivers. If I have to permit things first I could as well set up a CapabilityReceiver. So then I am not sure it is worth the effort personally.

Hmm; can you elaborate on what the `CapabilityReceiver` pattern looks like?

katelynsills:
> But this is the whole point of publish/claim, so far: "one-on-one easy transfer with verified identity"

I've been assuming we were trying to solve the bootstrapping problem that [Dete outlined](https://github.com/onflow/flow/pull/945#discussion_r954405266). The user problem is: you want to give a capability to someone else, and you don't want to give anyone other than the Recipient access to the capability. You have no prior connection to the Recipient. The current solutions have major flaws:

1. A co-signed transaction requires both signers to be online around the same time.
2. A dropbox-like approach would let someone other than the Recipient get access.

Here's how Dete [described the scenario](https://github.com/onflow/flow/pull/945#discussion_r954405266):

> I have a project that requires some amount of human oversight, maybe something as simple as a "pause" button that lets me block activity in the smart contract temporarily in response to a bug or market panic. As such, I have an administrator capability in my account that gives me access to call the "pause" function if necessary. In time, Bjarte joins my team, and I decide that it'd be useful if his account could also call the "pause" function in case I'm sleeping when the emergency happens. I'd like to send him a Capability for the same Admin object that I have. How do we get the Capability into his account securely?

The reason why Receivers don't work here is that they themselves fall prey to the bootstrapping problem. How does Bjarte give the receiver (a capability) specifically to Dete, and not anyone else? It's the same problem.

This FLIP really does a great job at solving this hard problem, in a way that doesn't get developers used to depending on msg.sender.

bluesign:
@katelynsills I think I explained this above like this:  ( this is what @bjartek refers as Capability Receiver pattern ) 
https://developers.flow.com/cadence/design-patterns#capability-receiver

```
this is already possible with:
0x2: set up account, put some resourceReceiver resource with function saveCapability(Capability)
0x2: set whitelist address: [0x1]
0x1: call saveCapability(Capability) with your capability (cap) ( 0x2: check if cap.address is in whitelist )
```

katelynsills:
> @katelynsills I think I explained this above like this: ( this is what @bjartek refers as Capability Receiver pattern ) 

Ah, good point @bluesign, thanks for the reminder!

> Problem with permit is, it is no different that capability receiver. You need account set up (permitted) before publishing.

But there is an important difference, right? We don't want smart contract developers to be writing code or using code at the smart contract level that is gated on identity. That doesn't allow delegation, and it will lead developers to not understand what capabilities can do that ACLs can't. 

If it's understood that `permit` is a platform functionality only for the initial bootstrapping, then we dodge that bullet. With `permit` no smart contract developer is using `msg.sender` patterns in their code.

bluesign:
> But there is an important difference, right? We don't want smart contract developers to be writing code or using code at the smart contract level that is gated on identity.

Yeah but if we limit this too much and not provide a benefit, people will continue using capability receiver and implement their ACL on top of that. 

We need to beat capability receiver pattern, to get usage and acceptance from developers. If we give something super restricted like this, I am afraid they will not use basically.  

> If I have to permit things first I could as well set up a CapabilityReceiver. 

We want people to drive slow, and then offering them slower car with worse interior.

katelynsills:
> Some analogy, we want people to drive slow, and then offering them slower car with worse interior.

@bluesign, that's a really good reminder that just saying "no" to developers isn't a good strategy. I'd be interested in your thoughts on how to engage other developers in seeing the really cool things that capabilities & safe, permissionless delegation can do. This FLIP seems like the wrong place for it, but that kind of discussion seems very valuable. I think allowing capabilities to have behaviors themselves and expanding beyond the facet pattern would help, but that's also outside the scope of this FLIP.

To continue your analogy, I think we should enforce a speed limit when we know it's a dead-end and there's a brick wall ahead (otherwise all our drivers are going to die in a fiery crash), but we also should tell the drivers that there's a high speed train right nearby that will let them go faster than any car would. And if the high speed train isn't built yet, we need to work on getting there :)

flips/20220908-publish-claim-capability.md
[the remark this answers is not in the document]
austinkline: Coming back to this

> Another thing to add here is that if events need to be emitted, the publishing party can always write a wrapper function in a smart contract, that publishes and emits an event.

Are you saying that events aren't necessary? Little confused by this. I specifically would advocate for a standardized event no matter what, the publisher should not be able to opt-out of that

> On the other hand, there is no way for the publisher to react when the value has been claimed. Maybe when publishing you could specify what happens when claiming? Some sort of callback. (that callback could emit an event or remove the published value)

A callback makes sense to me, otherwise the publisher would need to also listen to events and react to them in a separate transaction which doesn't feel right to me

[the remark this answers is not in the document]
dsainati1: Now we've settled on limiting this to just publishing/claiming capabilities, is a callback still necessary? `claim`ing a value automatically removes the publishing entry, so the publisher would not need to do that themselves. What is the use case (specifically as part of capability bootstrapping) for a callback here?

janezpodhostnik:
How do we currently do bootstrapping? We use [Capability Receivers](https://developers.flow.com/cadence/design-patterns#capability-receiver).

**A short rundown of that** (with **S** as the Sender a.k.a.: the flow account that wants to send/give a capability, and **R** the Receiver the flow account that is to receive this capability):
1. off-chain: S notifies R that it wants to send a capability to it (with a specific type!)
2. on-chain: R creates a capability receiver for that capability and exposes it publicly. R has to be careful with his implementation of the capability receiver since it is expose publicly and anyone can send a capability (potentially eating up space or confusing R on which capability is the right one)
3. off-chain: S becomes aware that R has created a receiver. This could happen by:
    - R telling S off chain
    - R triggering a designated on chain event
    - S pooling R for the existence of the receiver
4. on-chain: S sends the capability to R
5. on-chain: cleanup:
    - R removes the receiver
  
 **What are the problems:**
 
- This is a lot of steps
- Implementing a receiver correctly is not a given (and could be easily attacked if not correct). R might not know anything about receivers and how to correctly implemented and will just grab a recipe from somewhere, or worse, trust S to provide code on how to implement the receiver.
- S needs feedback on R being ready before it can proceed
 
**What I think the solution is**:

The original proposal had just `AuthAccount::Publish(value, name, account)` and `AuthAccount::Claim(name, address): value` (and of course you need unpublish as well), let me talk about that for a bit (and lets assume the value is a capability only). 

These functions change the bootstrapping pattern to:

1. on-chain: S publishes a capability for R (lets say by default no events get emitted) 
2. off-chain: S notifies R that it published a capability
3. on-chain: when R has time. R claims capability

Much simpler!

- by step 2. S is done
- before step 3. R doesn't have to do anything.
- S doesn't have to clean up, no-one else can claim the capability besides R (even if claiming wouldn't automatically un-publish it). S can clean up to save space (or if S has changed its mind about sharing at which point it should also un-link the capability just to be safe)
 - R does not have to write any custom code, it only has to store the capability somewhere, so It can use it later.
 
At this point there were questions about spamming and weather or not to send a common event when something is published. Can R be spammed, by constantly publishing for R? 

I don't think so. 
It's not like there is a notification popping up (on-chain) for R every-time someone publishes something for R. 
If S wants to let R know it published something S has to do that off-chain (and let the off-chain channels like gmail and slack deal with any spam :smiley:  ).
This is also the reason why there should **not** be a common chain event triggered when publishing. If there would be, people could rely on listening to them, at which point they could be spammed.
If S really wanted to it could still emit a (custom) event when publishing (maybe by wrapping the publish action in a smart contract function that also emits an event).

So in short I think something like this would be sufficient (I probably got the generic notation wrong) (these are all on AuthAccount):
```
fun publish<T: Any>(capability: Capability<T>, name: String, recipient: Address) // does not emit an event
fun unpublish(name: String, recipient: Address) 
fun claimFrom<T: Any>(name: String, provider: Address): Capability<T>? // does not un-publish
```

(I don't see the argument for needing to permit before something can be published

**(ab)use for other things than bootstrapping**

I don't think this is a  very efficient way of sending things, especially if we limit this to only capabilities.  The reason is that S must notify R off-chain that something is available to be picked up. If S and R intend to exchange a lot more stuff (than just one capability) maybe instead of going through bootstrapping each time, they should instead bootstrap a (pull) channel where S gives the capability to pull from the channel to R and for further sharing S just pushes more stuff to the channel. 

How that channel looks like is up to S and R, but I would imagine that S would want to emit an on-chain event that R knows about when adding to the channel.
 
 **Some other (random) thoughts**:
 - instead of publish/claim, wrap/unwrap could be used, where wrapping a capability into a `WrappedCapability<T>` makes it so that only the specified account can then unwrap it back into `Capability<T>`. This allows for the same pattern (of bootstrapping) but S had to provide its own way to serve `WrappedCapability<T>`s instead of publish providing this. On the other hand it might be more flexible, and allow for more patterns.
 - instead of publish/claim, we could use identity as in S can provide something to someone who can provide the identity of R (which is only R) but the problem with this is that the identity object would need to be sent into a function, and once it goes into a function S could capture it and pretend it is R.
 - If S knows for certain that R and only R can pick up the published capability, S can wrap the capability with a wrapper that basically says that "if anyone calls this R is to blame"

katelynsills:
If we decide for sure that no events are emitted, and the potential spam doesn't clog up block explorer displays, it makes sense to me to leave out `permit` as a step 👍

bjartek:
> Hmm; can you elaborate on what the `CapabilityReceiver` pattern looks like?

https://developers.flow.com/cadence/design-patterns#capability-receiver

bjartek:
> This is also the reason why there should **not** be a common chain event triggered when publishing. If there would be, people could rely on listening to them, at which point they could be spammed.
> If S really wanted to it could still emit a (custom) event when publishing (maybe by wrapping the publish action in a smart contract function that also emits an event).

I do not agree on this. I think a shared event should be emitted. Why is this "spamming" a problem here? Or is this something you claim? The benefits of emiting a shared event is numerous!

In .find we could listen to 1 event to know if a user has anything published to them, and then prompt them to accept it, or flow port could do the same. 

Shared events also signal that an mutation has happend on chain and that is a documented best practise. We should use more shared events, not less. Visbility is important.

pgebheim:
> If we decide for sure that no events are emitted, and the potential spam doesn't clog up block explorer displays, it makes sense to me to leave out `permit` as a step 👍 

I feel significantly less concerned about this capability spam than you seem to be.

This is not equivilent to token spam in ethereum because it doesn't create the same attack surface where a user further gives approval to their whole account to a malicious contract. Critically, the attack that's most concerning is actually having a token in your possession that looks valuable but is malicious.

If massive spam is a problem then it would mean that spammers would just be making themselves less visible, since there would be lots of noise in block explorers.

Even on networks with low fees, I have public accounts that have received some token spam but it's not THAT high. In this case the EV of the attack is so low that id be willing to ignore it so long as there's significant dev advantage to having a standard event.

dsainati1:
Even if we decide to not emit events, I still think allowing an open connection between any two accounts by means of this API does still pose an issue that @katelynsills mentioned earlier:

> For instance, imagine a motivated and coordinated harassment attack, where the Recipient keeps getting notifications where the string name is used to send threatening and abusive messages. Thus, there needs to be an easy way to revoke this connection between accounts.

If we don't want to require an initial `permit` step here in order to keep the happy path simple, there does still need to be some kind of `block` functionality or similar so that users do not have to be connected to abusive accounts if they do not wish to be.

pgebheim:
Does this actually need to be a protocol level concern?

Seems this type of blocking would be an application feature.

janezpodhostnik:
The accounts are not connected.
The Sender needs to notify the Recipient off-chain.

The only possible on chain connection they have, that would allow for abusive messaging is the common event emitted during publishing. 

I imagine that event would contain:
- a source address (Senders address)
- a destination address (Receivers address)
- the name of the provided value/capability
- possibly the type provided

If this event would exists I imagine people would want to listen to anything that was published for them, which opens them to receive spam (with a custom message in the name field). 
The workaround would be just listening to list of specific sources, but in this case I argue that its better to not have a common event and just let every sender define which event they are going to emit. If need be people can still opt to create a standard to follow where they would emit a certain event defined in a common contract somewhere. 

@bjartek @pgebheim  I still feel that there should not be a common (cadence-level) event emitted when publishing. But this is the only argument I have. 

I can perhaps suggest that we add the feature without the common event and open a separate discussion/PR for the common event.

bjartek:
> If this event would exists I imagine people would want to listen to anything that was published for them, which opens them to receive spam (with a custom message in the name field).

In my eyes this is a non-issue. There are numerous other ways a user can be spammed today. How many active users do not have a FLOAT account? A float user can be airdropped any NFT per the NFT standard. 

In Lost-And-Found you can add any resource or an NFT and send to a user and add a message. 

I have several more use cases where on chain "spam" is inevitable.


We need a common event or a way to onChain know if a user has something to claim for them.

katelynsills:
For what it's worth, here's a [real life example of the kind of abuse and harassment](https://www.dailymail.co.uk/news/article-11249985/Ex-boyfriend-repeatedly-cash-former-lovers-account-s-g-tramp-references.html) that I'm concerned about. Specifically, it's the name field that would be used to transmit a hateful message. An attack like this would prevent the recipient from being able to be use Flow at all.

One option that I haven't seen discussed yet is to keep the default open (anyone can publish to anyone, and every publish emits an event) but allow users to turn on an "Advanced Protection" mode in which they only accept published capabilities from accounts that they explicitly permit access. Seems like that could be the best of both worlds in terms of usability and security.

bjartek:
@katelynsills that is a really good example, and my point is that doing that is very easy already. And doing so in tools that is way easier available then this solution that is something that most likely only very technically savvy users/developers will user either way.

dsainati1:
> my point is that doing that is very easy already

I am not convinced that this is a particularly compelling argument; just because a bad practice is commonplace already does not excuse us as designers if we contribute to the problem by adding another vector for harassment.

dsainati1:
We had a public meeting about this FLIP today (Sept 26) and we settled on a few specific steps forward here:

1) Spam and harassment via events or messages is the responsibility of the user actor (e.g. the wallet) to manage, rather than the protocol, so we will remove the `allowlist`, `permit` and `unpermit` sections of this proposal in favor of these things being handled off-chain. We will likely suggest to wallet or other user actor developers that they do not display publishing events as-is to users, and allow users to restrict who they see events from. 

2) `publish`, `claim`, `unpublish` should all emit an event to signify that a capability has been published or removed, containing the name, type, and receiver (in the case of `publish`) and the name and removing account (in the case of `claim` and `unpublish`). This will give accounts the information they need to claim resources when they are published for them. 

3) For now, the API will continue to be restricted to capabilities. In the future we may decide to expand this to resources, but this would not be a breaking change (whereas going from resources now to capabilities in the future would be). 

4) `claim` should remove the capability being claimed from the publisher's dictionary. While capabilities can be duplicated, so it would not be semantically wrong to leave a copy present, leaving the capability in place would require manual cleanup on the part of the publisher that is not desirable. It would also encourage a pattern where receivers would repeatedly claim capabilities stored on the publisher's account whenever they need them, in order to avoid paying for the storage for that capability themselves.

janezpodhostnik (approved):

flips/20220908-publish-claim-capability.md
```diff
@@ -0,0 +1,149 @@
...
+of the value) calls `unpublish` with the same `name` that the value was originally stored with. If a value with that 
+`name` is present in the account's publishing dictionary, and the provided type argument to `unpublish` is a supertype
+of that value's runtime type, then the function will return that value and remove it from the dictionary. Otherwise, 
+the function returns `nil`. The type comparison is a force-cast like `load` and `claim`, so if the types do not match
+the program will fail. 
+
+This FLIP also adds two new events to Cadence:
+
+```cadence
+event InboxValuePublished(provider: Address, recipient: Address, name: String, type: Type)
```
turbolent, line 116: Is the naming of `Inbox` above and `InboxValue` intentionally generic/non-capability-specific to make it future-proof by preparing for a potential generalization from just capabilities to any value? If that's not the case, maybe make the naming more specific, i.e. `Inbox` -> `CapabilityInbox`, `InboxValuePublished` -> `CapabilityPublished`, etc.

turbolent (commented):
LGTM! Just a couple questions regarding naming (minor)

flips/20220908-publish-claim-capability.md
dsainati1: Yes; we had discussed in the meeting that we would limit it to capabilities for now with the possibility of widening later. In the interest of not having that widening be a breaking change if we decide it should happen, these names do not specify what kind of values are sent through the `Inbox`.

turbolent: Thank you for clarifying, that makes sense 👍

turbolent (approved):
Nice!

flips/20220908-publish-claim-capability.md
```diff
@@ -0,0 +1,149 @@
+# Publish/Claim Capability
+
+| Status        | (Proposed / Rejected / Accepted / Implemented)       |
```
dsainati1, line 3:
```suggestion
| Status        | (Accepted)       |
```

PR #1130, opened 2022-09-27
FLIP for a replacement for paths

janezpodhostnik:
## Description

Still in draft phase, but the idea is here.

### DIFF
--- a/flips/20220927-replacement-for-paths.md
+++ b/flips/20220927-replacement-for-paths.md
@@ -2,7 +2,7 @@
 
 | Status        | (**Draft** / Proposed / Rejected / Accepted / Implemented)       |
 :-------------- |:---------------------------------------------------- |
-| **FLIP #**    | [??](https://github.com/onflow/flow/pull/??)     |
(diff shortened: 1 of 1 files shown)


turbolent (commented):
Nice! 

An early version of Cadence had a non-path storage API which was keyed purely by types. I feel like a "compromise" like outlined in this proposal is probably going to solve the problems we encountered in the "only-types" (prototype) and "only paths" (current) APIs 👍

flips/20220927-replacement-for-paths.md
```diff
@@ -0,0 +1,89 @@
...
+    //     and one of them matches an empty string (the default)
+    //         return it
+    // otherwise
+    //     return null
+}
+```
+
+With this change there is less fear of collision between different daaps (because the types would be different) and the case of having only one thing of a certain type stored is the primary supported case. 
+
+!!TODO!! describe all API changes
```
turbolent, line 67: This section will likely need to describe how the changes will affect related APIs, such as storage iteration.

PR #1154, opened 2022-10-25
Add block data to execData API response

koko1123:
Updates the ExecDataByBlockID respoonse to contain the Block entity, and [mirror this object](https://github.com/onflow/flow-go/blob/da3188194739494eb013961ae885775255241554/engine/execution/computation/computer/uploader/model.go#L3-L22)


koko1123:
Not required as we can do. badger lookup

PR #1157, opened 2022-10-28
Add Block Status to Block Response

lolpuddle:
Closes: #3106

## Description

Adds block status as a field in the block response proto definition. (https://github.com/onflow/flow-go/issues/3106)

______

For contributor use:


koko1123 (approved):

protobuf/flow/entities/block.proto
```diff
@@ -12,6 +12,13 @@ import "flow/entities/block_seal.proto";
 import "flow/entities/execution_result.proto";
 import "flow/entities/block_header.proto";
 
+enum BlockStatus {
+  BLOCK_UNKNOWN = 0;
+  BLOCK_FINALIZED = 1;
+  BLOCK_EXECUTED = 2;
```
vishalchangrani, line 18: @lolpuddle when would a block be executed but not finalized or sealed?

lolpuddle: I guess realistically never because we'd probably return a storage error if we couldn't find the block, and if we were able to find the block locally then it would either be finalized, or sealed. I'll remove the finalized state

vishalchangrani (approved):

PR #1260, opened 2023-02-03
node operator guide to deal with a byzantine node

vishalchangrani:
Closes: #???

## Description

______

For contributor use:


docs/content/node-operation/byzantine-node-attack-response.mdx
```diff
@@ -0,0 +1,32 @@
...
+## Detection
+
+TODO
+
+## Action
+
+The response to an attack from a byzantine node is a three-pronged response.
+It entails first and foremost, blocking traffic coming from the byzantine node to your node.
+Then, raising a governance FLIP to remove the node from the network as described in this [FLIP](https://github.com/onflow/flips/blob/main/governance/20230105-identify-errant-node.md).
+And finally, a service account transaction to set the node weight to 0.
```
franklywatson, line 24:
```suggestion
Responding to an attack from a byzantine node requires the following: 

1. Immediate action to blocking network traffic routed from the byzantine node to your node.
2. Raising a governance FLIP to remove the node from the network as described in this [FLIP](https://github.com/onflow/flips/blob/main/governance/20230105-identify-errant-node.md).
3. A service account transaction to set the node weight to 0.
```

docs/content/node-operation/byzantine-node-attack-response.mdx
```diff
@@ -0,0 +1,32 @@
+---
+title: Byzantine node attack response
+sidebar_title: Byzantine attack response
+description: How to respond to a byzantine node attack on the network
+---
+
+Flow, like any blockchain, forms an open decentralized peer-to-peer network between all the nodes of the network.
+Hence, there exist an attack vector where one (or more than one) of the nodes is malicious (in other words byzantine) and intentionally tries to harm the network.
```
franklywatson, line 8:
```suggestion
Hence, there exists an attack vector where one (or more than one) of the nodes is malicious (in other words byzantine) and intentionally tries to harm the network.
```

docs/content/node-operation/byzantine-node-attack-response.mdx
```diff
@@ -0,0 +1,32 @@
...
+There exist several protections within the node software to deal with invalid messages - checking message signatures, checking message payload etc. are just some of them.
+These protections guard the network against any security attack. However, there could still be a byzantine node that spams the other nodes in the network with invalid data and slows them down.
+While this will not compromise the security of the network, it could hurt network liveliness.
```
franklywatson, line 11:
```suggestion
These protections guard the network against attacks. However, there could still be a byzantine node that spams the other nodes in the network with invalid data at volumes that are intended to impact node performance.
While this will not compromise the security of the network it could impact network liveness.
```

bluesign:
this seems very big power to give to node operator, without any consequences. What will happen node operator blocks a non-byzantine node as a byzantine one ?

vishalchangrani:
> this seems very big power to give to node operator, without any consequences. What will happen node operator blocks a non-byzantine node as a byzantine one ?
> 
> Is this only targeted for ANs? If so can we say we enable permissionless AN little premature?

If node operators block a byzantine node then their node will not receive all the data it needs to function. e.g. if a verification node blocks an execution node then it may not receive the execution results.

bluesign:
> If node operators block a byzantine node then their node will not receive all the data it needs to function. e.g. if a verification node blocks an execution node then it may not receive the execution results.

so this is only targeted for misbehaving AN ( byzantine is not apply here I guess ) ?

peterargue:
> so this is only targeted for misbehaving AN ( byzantine is not apply here I guess ) ?

This is intended for node operators to deal with attacks from other nodes on the network. Whether the behavior is intentional or a bug is hard to determine, so the triaging steps include coordinating with other operators to add context. I just added some more details to cover that part of the process.

bluesign:
> This is intended for node operators to deal with attacks from other nodes on the network. Whether the behavior is intentional or a bug is hard to determine, so the triaging steps include coordinating with other operators to add context. I just added some more details to cover that part of the process.

thanks for the clarification

peterargue (approved):

PR #1284, opened 2023-02-21
Clarifying on installation procedure

EmperorRP:
I was trying to follow the instructions during my setup of the SDK and it seemed slightly ambiguous, so this would be the perfect fix for the ambiguity.


bthaile (approved):
lgtm

PR #1289, opened 2023-02-23
Update header cards on Account Linking docs

chrisackermann:
Small corrections to header cards on Account Linking docs


sisyphusSmiling (approved):
LGTM!

PR #1301, opened 2023-03-08
[Access] Add start/end height to EventsResponse

peterargue:
## Description

Add start and end heights to the response from `GetEventsForHeightRange`. This allows clients to identify when the last available height on the Access node is lower than the `end_height` provided in the request


justinbarry:
ref https://github.com/onflow/flow-go/issues/4019

koko1123 (approved):

PR #1325, opened 2023-04-11
[Access] Add latest finalized block ID and height 

Guitarheroua:
https://github.com/onflow/flow-go/issues/2906

### Description

Extended necessary responses for the Access Node with additional fields containing information about the finalized block ID and height.


peterargue (commented):
Thanks for the PR @Guitarheroua!

This looks pretty good. 2 general comments:
* I think we should reframe the entity as `Metadata` so it can be extended in the future.
* Please also include the metadata in the following response objects:
  * `SendTransactionResponse`
  * `ExecuteScriptResponse`
  * `TransactionResultResponse`
  * `TransactionResultsResponse` (we'll want to omit the metadata within each `TransactionResultResponse` when sending this)

protobuf/flow/entities/metadata.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+package flow.entities;
+
+option go_package = "github.com/onflow/flow/protobuf/go/flow/entities";
+option java_package = "org.onflow.protobuf.entities";
+
+message Metadata {
+  bytes latest_finalized_block_id = 1;
+  uint64 latest_finalized_height = 2;
```
peterargue, line 10: I should have asked in my previous request. Can you also add `bytes node_id`

peterargue (approved):

koko1123 (approved):

PR #1328, opened 2023-04-12
Revert "[Access] Add start/end height to EventsResponse"

koko1123:
Reverts onflow/flow#1301


peterargue (approved):

peterargue:
I agree this isn't ready to merge. There are backwards compatibility issues with the REST implementation, so I think we may need to take a different approach.

PR #1345, opened 2023-06-09
Update sporks.json for testnet45 and testnet46

koko1123:


zhangchiqing (approved):
agree with peter's comments

PR #1275, opened 2023-02-10
Add streaming endpoints to executiondata api

peterargue:
Closes: #???

## Description

______

For contributor use:

- [ ] Targeted PR against `master` branch
- [ ] Linked to Github issue with discussion and accepted design OR link to spec that describes this work.
- [ ] Updated relevant documentation 
- [ ] Re-reviewed `Files changed` in the Github PR explorer
- [ ] Added appropriate labels

PR #1353, opened 2023-07-22
Fix/executiondata proto pkg

treethought:
Closes: #???

## Description

Fixes the package name for executiondata protobuf file from access -> executiondata.
Related to https://github.com/onflow/flow/pull/1275 and https://github.com/onflow/flow-go-sdk/issues/378 and needed for https://github.com/onflow/flow-go-sdk/pull/417

Previously, with package `access`, the `executiondata` types could not be imported in generated go code.

______

For contributor use:


koko1123:
Hey this is a good idea because of the `access` package name overlap, but the intent behind the execution data APi was to be a subset of the Access API. The generated go files will break the `flow-go` implementation, so we have two possible routes around it:
1. Modify the corresponding handlers in `flow-go` to avoid copying mutex values
2. re-generate the go files with the older version of `buf` (in go.mod of this repo)

I suggest making a PR in `flow-go` for 1., then merging this PR, then updating the `flow-go` PR with the merged PR hash as a dependency

protobuf/README.md
```diff
@@ -11,6 +11,7 @@ Check out the [Flow Access API specification](/docs/content/access-api.md).
 You can use [buf](https://github.com/bufbuild/buf) to generate gRPC client stubs in a variety of languages.
 Please make sure you have `protoc-gen-go-grpc` installed, for example using command
 ```shell script
+go install google.golang.org/protobuf/cmd/protoc-gen-go@v1.28
```
peterargue, line 14: I'm not opposed to making this change, but I think we should do it as a separate PR

treethought: I had added that line for what I used since I couldn't find what version was being used, but just found via https://github.com/onflow/flow/pull/905#issuecomment-1097221140 that it seems to be set in the flow-go makefile as [v1.3.2 ](https://github.com/onflow/flow-go/blob/master/Makefile#L84)

peterargue (approved):

protobuf/README.md
peterargue: ah ok. thanks for fixing the docs!

peterargue:
FYI @nvdtf, @Guitarheroua

PR #1382, opened 2023-09-12
[Access API] Add transaction results to execution data entities

peterargue:
## Description

Add `TransactionResults` to `ChunkExecutionData` entities within the execution data API

See: https://github.com/onflow/flow-go/issues/4707


protobuf/flow/entities/block_execution_data.proto
```diff
@@ -8,33 +8,98 @@ option java_package = "org.onflow.protobuf.entities";
...
+  // List of transactions included in the collection.
   repeated Transaction transactions = 1;
 }
 
+// TrieUpdate produced by executing the collection.
+//
+// TrieUpdates contain a list of registers that were modified during chunk execution. The value
+// included is the new value of the register.
 message TrieUpdate {
+  // RootHash is the root hash of the trie that included the contained updates.
```
zhangchiqing, line 61:
```suggestion
  // RootHash is the root hash of the trie that before the update is applied.
```

protobuf/flow/entities/block_execution_data.proto
```diff
@@ -8,33 +8,98 @@ option java_package = "org.onflow.protobuf.entities";
...
   repeated bytes paths = 2;
+
+  // List of register values updated.
+  //
+  // Note: paths and payloads map 1:1 with eachother. i.e. for each element in path, the value in
+  // payloads at the same index is the value of the register at that path.
   repeated Payload payloads = 3;
 }
 
+// Payload represents the value of a register.
```
zhangchiqing, line 77:
```suggestion
// Payload represents the key-value pair of a register
```

protobuf/flow/entities/block_execution_data.proto
```diff
@@ -8,33 +8,98 @@ option java_package = "org.onflow.protobuf.entities";
...
 message Payload {
+  // List of key parts that make up the register key.
```
zhangchiqing, line 79:
```suggestion
  // List of key parts that make up the register key. can be converted into register id
```

protobuf/flow/entities/block_execution_data.proto
zhangchiqing: See the implementation here, where a ledger update [is created with State as baseState](https://github.com/onflow/flow-go/blob/v0.31/engine/execution/state/state.go#L237). (baseState is the root hash of the trie before the update).

[And here the ledger update is converted into trie update with state copied to RootHash field](https://github.com/onflow/flow-go/blob/v0.31/ledger/complete/ledger.go#L206)

koko1123 (approved):
Thanks for the documentation!

PR #1391, opened 2023-10-11
[Access] 4506 - Make CCF encoded events optional via request parameter

Guitarheroua:
## Description

https://github.com/onflow/flow-go/issues/4506

This pull request introduces a new request parameter for a list of endpoints from the [AccessAPI](https://github.com/onflow/flow-go/blob/master/access/api.go) and the [ExecutionDataAPI](https://github.com/onflow/flow-go/blob/dafd8fd29c476cacc497f79940412ba377f723fe/engine/access/state_stream/backend.go#L73), which specifies the encoding for events.


protobuf/flow/executiondata/executiondata.proto
```diff
@@ -67,6 +67,7 @@ service ExecutionDataAPI {
 message GetExecutionDataByBlockIDRequest {
   // Block ID of the block to get execution data for.
   bytes block_id = 1;
+  entities.EventEncodingVersionValue event_encoding_version = 2;
```
peterargue, line 70: please add comments for this parameter similar to other parameters in this fail

peterargue (approved):
Please update the comments for the execution data api to align with the other arguments, but otherwise this looks good.

PR #1404, opened 2023-11-15
[Execution Data] Change registerID definition to match Ledger.RegisterID

koko1123:
## Description
The existing protobuf requires registerID in the `[][]byte` format, which breaks round trip conversion. I have made a new field in the request entity, `register_ids_v2` to match the `flow-go` `ledger.RegisterID`


protobuf/flow/executiondata/executiondata.proto
```diff
@@ -221,7 +222,10 @@ message GetRegisterValuesRequest {
   uint64 block_height = 1;
 
   // Register IDs of the Ledger.RegisterID format with an owner and key.
-  repeated bytes register_ids = 2;
+  repeated bytes register_ids = 2; // deprecated!
+
+  // Register IDs of the Ledger.RegisterID format with an owner and key.
+  repeated  entities.RegisterID register_ids_v2 = 3;
```
peterargue, line 228: since this proto was never implemented, I think it's fine to just update the original
```suggestion
  // Register IDs of the Ledger.RegisterID format with an owner and key.
  repeated  entities.RegisterID register_ids = 2;
```

peterargue (approved):

PR #1433, opened 2024-02-16
[Access] Implement gRPC streaming endpoint SubscribeAccountStatuses

AndriiDiachuk:
In this PR was Introduced changes in proto for a new endpoint named SubscribeAccountStatuses.


protobuf/flow/executiondata/executiondata.proto
```diff
@@ -230,4 +254,78 @@ message GetRegisterValuesResponse {
 
   // raw register values at the given height.
   repeated bytes values = 1;
-}
\ No newline at end of file
+}
+
+// The request for SubscribeAccountStatuses
+message SubscribeAccountStatusesRequest {
```
durkmurder, line 260: Since we are introducing new API I suggest to enforce "having only `start_block_id` or `start_block_height`" in protobuf using [oneof](https://protobuf.dev/programming-guides/proto3/#oneof) keyword. This will make a cleaner API and simplify implementation.

protobuf/flow/executiondata/executiondata.proto
```diff
@@ -230,4 +254,78 @@ message GetRegisterValuesResponse {
...
+  bytes address = 2;
+
+  // Events matching the StatusFilter in the request.
+  // The API may return no events which signals a periodic heartbeat. This
+  // allows clients to track which blocks were searched. Client can use this
+  // information to determine which block to start from when reconnecting.
+  repeated entities.Event events = 3;
+
+  // The MessageIndex of the response message.
+  // Used by the client to ensure they received all messages.
```
durkmurder, line 312:
```suggestion
  // Used by the client to ensure they have received all messages.
```

protobuf/flow/executiondata/executiondata.proto
peterargue: instead of using `oneof`, let's split the API into 3 methods like the polling endpoints. I made a similar request for the block streaming endpoints. e.g.
* `SubscribeAccountStatusesFromStartBlockID`
* `SubscribeAccountStatusesFromStartHeight`
* `SubscribeAccountStatusesFromLatestBlock`

This adds a bit of boiler plate, but simplifies the client and server side logic

protobuf/flow/executiondata/executiondata.proto
```diff
@@ -230,4 +302,135 @@ message GetRegisterValuesResponse {
...
+        // information to determine which block to start from when reconnecting.
+        repeated entities.Event events = 2;
+  }
+  repeated Result results = 4;
+}
+
+// StatusesFilter defines the filter to apply to block events.
+// Filters are applied as an OR operation, i.e. any event matching any of the
+// filters is returned. If no filters are provided, all events are returned. If
+// there are any invalid filters, the API will return an InvalidArgument error.
```
peterargue, line 415:
```suggestion
// Filters match for events with types in the included event_type list, that are
// related to at least one address from the provided address list. An event who's
// type matches but address does not is ignored, and vice versa. 
// If no event_types are provided, all account related protocol event types are matched.
// If no addresses are provided, any address matches.
// If there are any invalid filters, the API will return an InvalidArgument error.
```

protobuf/flow/executiondata/executiondata.proto
```diff
@@ -230,4 +302,135 @@ message GetRegisterValuesResponse {
...
+  message Result {
+      // Unique identifier for the account being streamed
+        bytes address = 1;
+        // Events matching the StatusFilter in the request.
+        // The API may return no events which signals a periodic heartbeat. This
+        // allows clients to track which blocks were searched. Client can use this
+        // information to determine which block to start from when reconnecting.
+        repeated entities.Event events = 2;
+  }
+  repeated Result results = 4;
```
peterargue, line 409: this comment should apply to the results
```suggestion
  message Result {
        // Unique identifier for the account being streamed
        bytes address = 1;
        // Events matching the StatusFilter in the request.
        repeated entities.Event events = 2;
  }
  // The API may return no results which signals a periodic heartbeat. This
  // allows clients to track which blocks were searched. Client can use this
  // information to determine which block to start from when reconnecting.
  repeated Result results = 4;
```

protobuf/flow/executiondata/executiondata.proto
```diff
@@ -230,4 +302,135 @@ message GetRegisterValuesResponse {
...
+  // 2. JSON-CDC
+  entities.EventEncodingVersion event_encoding_version = 3;
+}
+
+// The response for SubscribeAccountStatuses
+message SubscribeAccountStatusesResponse {
+  // Block ID of the block containing the events.
+  bytes block_id = 1;
+
+  // Block ID of the block containing the events.
```
peterargue, line 393:
```suggestion
  // Block height of the block containing the events.
```

protobuf/go/flow/access/access_grpc.pb.go
```diff
@@ -1,6 +1,6 @@
 // Code generated by protoc-gen-go-grpc. DO NOT EDIT.
 // versions:
-// - protoc-gen-go-grpc v1.3.0
+// - protoc-gen-go-grpc v1.2.0
```
peterargue, line 3: looks like you generated with an older version than the last person to update these. can you update your client?

peterargue (commented):
added a few comments about comments. but the structure looks good.

peterargue (commented):
looks good, but can you upgrade your `protoc-gen-go-grpc` to `v1.3.0`. looks like it regenerated a few unrelated protobufs because your version is out of date.

peterargue (approved):

PR #1451, opened 2024-04-16
[Access] Add new rpc endpoints

illia-malachyn:
https://github.com/onflow/flow-go/issues/5558

Added the following rpc endpoints:
- SubscribeExecutionDataFromStartBlockID
- SubscribeExecutionDataFromStartBlockHeight
- SubscribeExecutionDataFromLatest


protobuf/flow/executiondata/executiondata.proto
```diff
@@ -38,8 +38,46 @@ service ExecutionDataAPI {
   rpc SubscribeExecutionData(SubscribeExecutionDataRequest)
       returns (stream SubscribeExecutionDataResponse);
```
peterargue, line 39: please also add a deprecation notice similar to what was added for the old streaming events endpoint
https://github.com/onflow/flow/blob/e4cf6a528d3c848312eb7be959a9e8f6a580a5a7/protobuf/flow/executiondata/executiondata.proto#L66-L69

peterargue (approved):

PR #1452, opened 2024-04-18
Andrii slisarchuk/fix openapi for documentation

Guitarheroua:
https://github.com/onflow/docs/issues/464

This is a fix for the OpenAPI, that does not display correctly the `access.yaml` WebSocket errors in [documentation](https://github.com/onflow/docs/blob/main/docusaurus.config.js#L219).

### DIFF
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -573,7 +573,10 @@ paths:
   /subscribe_events:
     get:
       summary: Subscribe events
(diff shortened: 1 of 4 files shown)


peterargue (approved):
thanks for fixing this. The response code scheme you used seems like a reasonable work around to openapi not supported websockets.

PR #1475, opened 2024-06-18
Added new methods GetAccountBalance* and GetAccountKeys*

AndriiDiachuk:
In this PR two new calls `GetAccountBalance`* and `GetAccountKeys`* were added to proto.

Related issue: https://github.com/onflow/flow-go/issues/5999


protobuf/flow/access/access.proto
```diff
@@ -424,6 +446,42 @@ message GetAccountAtBlockHeightRequest {
...
+  entities.Metadata metadata = 2;
+}
+
+message GetAccountKeysRequest {
+  bytes address = 1;
+}
+
+message GetAccountKeysAtLatestBlockRequest {
+  bytes address = 1;
+}
```
peterargue, line 473: Let's also provide support for specifying a specific key index, which is something currently supported by the REST API. since this field will be a number whose default (0) is a valid index, we'll need to wrap it in another message to allow the server to differentiate 0 and unset.

something like
```suggestion
message GetAccountKeysAtLatestBlockRequest {
  // address of account 
  bytes address = 1;
  // index of key to return
  // omit to return all keys
  entities.AccountKeyIndex index = 2;
}
```

then you could define this in https://github.com/onflow/flow/blob/master/protobuf/flow/entities/account.proto
```golang
message AccountKeyIndex {
  uint32 value = 1;
}
```

protobuf/flow/access/access.proto
```diff
@@ -424,6 +451,55 @@ message GetAccountAtBlockHeightRequest {
...
+message GetAccountKeysAtLatestBlockRequest {
+  // address of account
+  bytes address = 1;
+}
+
+message GetAccountKeyAtLatestBlockRequest {
+  // address of account
+  bytes address = 1;
+  // index of key to return
+  entities.AccountKeyIndex index = 2;
```
peterargue, line 478: since we're using separate endpoints for `*Key` vs `*Keys`, you can flatten this to just a `uint32`. It will simplify the logic for clients.

protobuf/flow/access/access.proto
```diff
@@ -424,6 +451,55 @@ message GetAccountAtBlockHeightRequest {
...
+}
+
+message GetAccountBalanceAtBlockHeightRequest {
+  bytes address = 1;
+  uint64 block_height = 2;
+}
+
+message AccountBalanceResponse {
+  uint64 balance = 1;
+  uint64 available_balance = 2;
```
peterargue, line 465: as noted in the flow-go PR, let's remove `available_balance` for now. we can add it back later.

peterargue (approved):

peterargue (approved):

PR #1491, opened 2024-07-23
[Access] Add REST endpoint to get an accounts flow balance

AndriiDiachuk:
In this PR the openAPI spec was updated with new GetAccountBalance call.

### DIFF
--- a/openapi/access.yaml
+++ b/openapi/access.yaml
@@ -399,9 +399,7 @@ paths:
           schema:
             $ref: '#/components/schemas/BlockHeight'
           required: false
-          description: The block height to query for the account details at the "sealed" is used by default.
-        - $ref: '#/components/parameters/expandParam'
-        - $ref: '#/components/parameters/selectParam'
+          description: The block height to query for the account balance. "sealed" is used by default.
       responses:
         '200':
           description: OK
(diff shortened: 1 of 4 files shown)


openapi/access.yaml
```diff
@@ -381,6 +381,40 @@ paths:
...
+          required: true
+          description: The address of the account.
+        - name: block_height
+          in: query
+          schema:
+            $ref: '#/components/schemas/BlockHeight'
+          required: false
+          description: The block height to query for the account details at the "sealed" is used by default.
+        - $ref: '#/components/parameters/expandParam'
+        - $ref: '#/components/parameters/selectParam'
```
peterargue, line 404: you can remove these since there's nothing to select or expand in the response
```suggestion
```

openapi/access.yaml
```diff
@@ -381,6 +381,40 @@ paths:
...
+          schema:
+            $ref: '#/components/schemas/Address'
+          required: true
+          description: The address of the account.
+        - name: block_height
+          in: query
+          schema:
+            $ref: '#/components/schemas/BlockHeight'
+          required: false
+          description: The block height to query for the account details at the "sealed" is used by default.
```
peterargue, line 402:
```suggestion
          description: The block height to query for the account balance. "sealed" is used by default.
```

PR #1496, opened 2024-08-05
 [Access] Add compatible range to Node Version Info endpoint

UlyanaAndrukhiv:
## Context

In this PR the `openAPI` and `protobuf` spec was updated for `NodeVersionInfo` endpoint.


protobuf/flow/entities/node_version_info.proto
```diff
@@ -21,4 +21,8 @@ message NodeVersionInfo {
     // If the node started after the beginning of the spork, it is the height of the first sealed block
     // indexed.
     uint64 node_root_block_height = 6;
+    // The first block that the node's version supports.
+    uint64 protocol_version_start_height = 7;
+    // The last block the node's version supports.
+    uint64 protocol_version_end_height = 8;
```
peterargue, line 27: I think these should be grouped in an embedded message. that way if version control is not enabled, we can just return nil indicating there are no values.

also note that this is not the protocol version start/end. It's actually the execution version since it's related to cadence/fvm. there will be a future change to add support for similar functionality for different protocol versions, so we should make sure to differentiate them here.

how about naming the message field `compatible_range` with `start_height` and `end_height` fields

openapi/access.yaml
```diff
@@ -1218,6 +1220,12 @@ components:
         node_root_block_height:
           type: string
           format: uint64
+        protocol_version_start_height:
+          type: string
+          format: uint64
+        protocol_version_end_height:
+          type: string
+          format: uint64
```
peterargue, line 1228: similar comment for the rest API

peterargue (approved):

PR #1539, opened 2025-01-13
[Access] Update model Block REST API Response

UlyanaAndrukhiv:
issue: https://github.com/onflow/flow-go/issues/6874

## Context

 In this PR the `openAPI` was updated: 
 - the `omitempty` tag was added to the `Expandable` field for `Block`,
 - the `block_status` field was added for `Block`


openapi/access.yaml
```diff
@@ -955,6 +955,8 @@ components:
               format: uri
         _links:
           $ref: '#/components/schemas/Links'
+        block_status:
```
UlyanaAndrukhiv, line 958: The `block_status` field was introduced in `flow-go`'s `model_block.go` but was missing here. This addition ensures consistency between the two models.

For reference, the commit where `block_status` was added to flow-go: [commit](https://github.com/onflow/flow-go/commit/f89fef072a9bad617639970be81838c100211d5b)

openapi/go-client-generated/api/swagger.yaml
```diff
@@ -1591,7 +1591,10 @@ components:
           $ref: '#/components/schemas/Block__expandable'
         _links:
           $ref: '#/components/schemas/Links'
+        block_status:
+          type: string
```
peterargue, line 1595: Can you make this an enum like `TransactionExecution` and `TransactionStatus`?

openapi/access.yaml
```diff
@@ -933,11 +933,18 @@ components:
         - Executed
         - Sealed
         - Expired
+    BlockStatus:
+      type: string
+      description: This value indicates the status of a block.
+      enum:
+        - BLOCK_UNKNOWN
+        - BLOCK_FINALIZED
+        - BLOCK_SEALED
```
peterargue, line 942: for reference, these are defined [here](https://github.com/onflow/flow-go/blob/1ba41690954621d15d59dcabcc8f0dfc0b56f1bd/model/flow/block.go#L61-L75)

peterargue (approved):

PR #1573, opened 2025-04-07
Remove reference to Canary network from Swagger documentation

sjonpaulbrown:
## Description
We are deprecating the Canary network, and the following PR removes the reference from the `access.yaml` file and includes the re-generated swagger docs.


vishalchangrani (approved):
thanks a lot for this cleanup JP!!
Often this is the kind of stuff that lingers forever so thanks for cleaning it out.

PR #1613, opened 2025-07-25
add ChainID field to CollectionGuarantee

tim-barry:
## Description

Adds the `ChainID` field of CollectionGuarantees to the exported gRPC API and REST API (renamed to `cluster_chain_id`, to avoid confusion with the chain ID of the main network).
This is so that the ID of a CollectionGuarantee (and therefore integrity of the chain) can be calculated by users of the API (notably, Rosetta) after the Malleability changes, required by https://github.com/onflow/flow-go/issues/7630

______

For contributor use:


protobuf/flow/entities/collection.proto
```diff
@@ -17,4 +17,5 @@ message CollectionGuarantee {
   bytes signature = 4;
   repeated bytes signer_ids = 5; // deprecated!! value will be empty. replaced by signer_indices
   bytes signer_indices = 6;
+  bytes chain_id = 7;
```
peterargue, line 20: this is the cluster chain id right? How about naming it `cluster_chain_id` so users don't confuse it with the network's `chain_id`

jordanschalm (approved):

PR #1622, opened 2025-08-22
Add execution state query field to executiondata protobuf

illia-malachyn:
Issue: https://github.com/onflow/flow-go/issues/7759

This is a part of "Data Avaliability" project.
I added an execution data query arg for execution data models as well as executor metadata for their responses.

I formatted the models in a separate PR https://github.com/onflow/flow/pull/1623


protobuf/flow/executiondata/executiondata.proto
```diff
@@ -468,6 +481,9 @@ message SubscribeEventsResponse {
 
   // The message index of the response message. Used by the client to ensure they received all messages. Starts from "0".
   uint64 message_index = 5;
+
+  // Metadata about the execution result that was used to produce this response.
+  entities.ExecutorMetadata executor_metadata = 6;
```
illia-malachyn, line 486: @peterargue Do we need to include executor metadata for these endpoints ?

peterargue (approved):

PR #1626, opened 2025-08-27
rename metadata to metadata

illia-malachyn:
rename `execution_state_query` to `executor_metadata`


illia-malachyn:
Added leftover changes to Andrii's PR

